PR #1, opened 2021-06-01
Define Livebook rendering protocol

jonatanklosko:
See https://github.com/elixir-nx/livebook/pull/309.


lib/kino/livebook_output.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Kino.LivebookOutput do
+  @moduledoc """
+  A number of output formats supported by Livebook.
+  """
+
+  @typedoc """
+  Livebook cell output may be one of these values
+  and is rendered accordingly.
+  """
+  @type t ::
```
jonatanklosko, line 10: @josevalim we discussed using a `%Kino.Output{}` struct, but I realised the output type must be recognisable in Livebook itself, so I think it's ok to stick to tuples (and this one-of type is also neat for the plain string output).

lib/kino/livebook_output.ex
```diff
@@ -0,0 +1,58 @@
...
+          :ignored
+          # Regular text, adjacent such outputs can be treated as a whole
+          | binary()
+          # Standalone text block
+          | {:text, binary()}
+          # Vega-Lite graphic
+          | {:vega_lite_static, spec :: map()}
+          # Vega-Lite graphic with dynamic data
+          | {:vega_lite_dynamic, widget_process :: pid()}
```
jonatanklosko, line 19: @josevalim One thing I'm not quite certain about is where to document the communication protocol. This output type is theoretically Kino-agnostic, because anyone could have a process that exposes Vega-Lite stuff, but currently the protocol is described [in kino](https://github.com/elixir-nx/kino/blob/22411b82e2d5a3d12d7c5738c86f01e958fc0257/lib/kino/vega_lite.ex#L25-L37).

lib/kino/transform.ex
```diff
@@ -0,0 +1,36 @@
+defprotocol Kino.Transform do
```
jonatanklosko, line 1: `Transformable`, something else?

lib/kino/livebook_output.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Kino.LivebookOutput do
```
josevalim, line 1: Not sure I would call it LivebookOutput. One of the paths we might explore in the future is to render the same output in different media. So, in order to not paint ourselves into a corner, I would call this `Kino.Output`.

lib/kino/transform.ex
josevalim: I like `Kino.Render`.

lib/kino/widget.ex
```diff
@@ -1,30 +1,21 @@
...
-          type: atom()
-        }
+  @moduledoc false
 
   @doc false
-  @spec start!(module(), keyword(), atom()) :: Kino.Widget.t()
-  def start!(module, opts, type) do
+  @spec start!(module(), keyword()) :: pid()
+  def start!(module, opts) do
     case DynamicSupervisor.start_child(Kino.WidgetSupervisor, {module, opts}) do
```
josevalim, line 7: I would call this module for now. You might as well call `DynamicSupervisor.start_child(Kino.WidgetSupervisor, {module, opts})` directly in `Kino.VegaLite` (and you know it can only succeed with `{:ok, pid}` anyway).

lib/kino/livebook_output.ex
jonatanklosko: Yeah, but this is specific to `to_livebook` and the functions return Livebook-compatible outputs. If we were to output something different then we would add `to_text` function to the `Kino.Render` protocol, but that wouldn't use this module at all, right?

josevalim: That's the thing, I don't know yet... I can see our render output be generic enough to render in multiple places. For example, I can see a table() working on both livebook and also other static scenarios. So I would go with a generic name just in case.

If you want to remain consistent, you can also rename the `to_livebook` function to something generic too, such as `to_output`. And then we can add specializations later on (such as `to_livebook`, `to_console`, etc). WDYT?

josevalim: Actually, now that I wrote it, I like `Kino.Output` and `Kino.Render.to_output` a lot. WDYT?

jonatanklosko: It just kinda covers the fact that the output is Livebook-compatible, but sounds alright to me. One thing to keep in mind is that Livebook will depend on the function name so changing it later may be awkward.

I guess your point is that if we have `to_console` later on, it could still call `to_output` and then render that output on its own, rather than working with the plain value?

lib/kino/widget.ex
jonatanklosko: You mean it should be fine to start and aggressively match against `{:ok, pid}` in `Kino.VegaLite`?

josevalim: Yes. Do you know any reason why it should fail?

lib/kino/livebook_output.ex
josevalim: I was thinking `to_output` would always work and be the general fallback, but we would allow specializations on `to_livebook` and `to_console` or similar?

lib/kino/widget.ex
jonatanklosko: I don't :) And in cases like not enough memory it doesn't really matter since things are bad enough already, right?

lib/kino/livebook_output.ex
jonatanklosko: So for me `Kino.Output` should return formats compatible with Livebook, which means `to_output` and `to_livebook` are essentially the same.

If we have a conceptually generic `to_output` then the difference would likely lie in `Kino.render` rather than the protocol implementation right?

josevalim: Sure. I think Kino.Output with Kino.Render.to_livebook works fine too!

jonatanklosko: Actually, if we assume that every `Kino.Output.t()` is a valid `Livebook.Notebook.Cell.output()`, then this looks quite nice:

```elixir
defmodule Livebook.Evaluator.DefaultFormatter do
  ...

  defp to_output(value) do
    # Kino is a "client side" extension for Livebook that may be
    # installed into the runtime node. If it is installed we use
    # its more precies output rendering rules.
    if Code.ensure_loaded?(Kino.Render) do
      try do
        Kino.Render.to_output(value)
      catch
  ...
```

lib/kino/livebook_output.ex
jonatanklosko: At least from documentation perspective, we could describe the protocol under a separate type like this:

```elixir
  @typedoc """
  Livebook cell output may be one of these values
  and is rendered accordingly.
  """
  @type t ::
          :ignored
          # Regular text, adjacent such outputs can be treated as a whole
          | binary()
          # Standalone text block
          | {:text, binary()}
          # Vega-Lite graphic
          | {:vega_lite_static, spec :: map()}
          # Vega-Lite graphic with dynamic data
          | {:vega_lite_dynamic, vega_lite_pid()}

  @typedoc """
  A process wrapping VegaLite graphic and allowing to
  dynamically send data to the clients.

  ## Communication protocol

  A client process should connect to the VegaLite process by sending:

      {:connect, pid}

  And expect the following reply:

      {:connect_reply, %{spec: map()}}

  The VegaLite process may then keep sending one of the following events:

      {:push, %{data: list(), dataset: binary(), window: non_neg_integer()}}
  """
  @type vega_lite_pid :: pid()
```

I'm not sure if this should live here on in `Livebook.Notebook.Cell.output` though. Thoughts?

josevalim: It is all "private" for now, so I think it is fine to document it here. You can also do this:

```elixir
| vega_lite_static()
```

And then:

```elixir
@doc ...
@type vega_lite_static() :: {:vega_lite_static, spec :: map()}
```

lib/kino/livebook_output.ex
josevalim: It is your call. Both `to_output`/`to_livebook` are fine by me! Go with `to_livebook` if you want to be conservative.

josevalim (approved):

lib/kino/livebook_output.ex
jonatanklosko: I documented all types in this manner, hopefully not overengineered :see_no_evil:

PR #6, opened 2021-06-15
Change docs main page to README

wojtekmach:
I think examples is a great starting point but given it's common to start with a README, there is the license information, and we have the convenient "Next" button to go to examples, I thought I'd open it up. But yeah, happy to leave things as is, just curious :)


jonatanklosko (approved):
Sounds good to me, especially now that we have this lovely logo, thanks!

PR #7, opened 2021-06-16
Return the Vl widget on periodically

josevalim:
This way, in Livebook, instead of

    widget =
      Vl.new(...)
      |> ...
      |> Kino.VegaLite.start()
      |> Kino.render()

    Kino.VegaLite.periodically(widget, ...)

we can do

    Vl.new(...)
    |> ...
    |> Kino.VegaLite.start()
    |> Kino.VegaLite.periodically(...)

Should we do the same for push/clear/etc?

### DIFF
--- a/lib/kino/vega_lite.ex
+++ b/lib/kino/vega_lite.ex
@@ -112,9 +112,10 @@ defmodule Kino.VegaLite do
 
   The callback is run for the first time immediately upon registration.
   """
-  @spec periodically(t(), pos_integer(), term(), (term() -> {:cont, term()} | :halt)) :: :ok
+  @spec periodically(t(), pos_integer(), term(), (term() -> {:cont, term()} | :halt)) :: t()
   def periodically(widget, interval_ms, acc, fun) do
     GenServer.cast(widget.pid, {:periodically, interval_ms, acc, fun})
+    widget
   end
 
   @impl true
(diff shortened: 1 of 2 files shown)


jonatanklosko:
So this is convenient, but my main concern was such scenario:

**Cell 1**

```elixir
widget =
  Vl.new(...)
  |> ...
  |> Kino.VegaLite.start()
```

**Cell 2**

```elixir
Kino.VegaLite.periodically(widget, ...)
```

**Cell 3**

```elixir
Kino.VegaLite.push(widget, ...)
```

I think we should render the plot only under the first cell. But perhaps that shouldn't stop us from going the convenience route?

> Should we do the same for push/clear/etc?

Either way, for these operations I don't think we should return the widget (I don't think it makes anything easier, whereas there is the case above).

josevalim:
For me the periodically is part of the specification rather than part of the push, so I would most likely have it with cell 1, if that makes any sense.

jonatanklosko:
> For me the periodically is part of the specification rather than part of the push, so I would most likely have it with cell 1.

Makes sense, fwiw I don't think having two separate cells like that should be a common scenario, so streamlining the usual case sounds good.

Could you update the example notebook to omit the additional render?

josevalim:

josevalim:
> Could you update the example notebook to omit the additional render?

Will do whenever there is a new Kino version out (although no rush on that).

PR #8, opened 2021-06-16
Update introductory docs

josevalim:


mix.exs
```diff
@@ -35,13 +35,10 @@ defmodule Kino.MixProject do
 
   defp docs do
     [
-      main: "readme",
```
jonatanklosko, line 38: Will miss the logo :see_no_evil: Should we have a text-less version of the logo in the sidebar?

josevalim: Yes, let's do that instead :) Can you please do so in a separate commit?

jonatanklosko: Surething!

jonatanklosko (approved):

josevalim:

PR #12, opened 2021-06-16
Add dynamic table output and ETS table widget

jonatanklosko:
This defines a new output type - an interactive data table. The client (Livebook) makes queries to the widget process to get rows (with optional support for pagination and sorting). For the output rendering see https://github.com/elixir-nx/livebook/pull/356.

Further, this defines `Kino.ETS` widget that adheres to the dynamic table communication protocol and makes for an interface for browsing an ETS table.

https://user-images.githubusercontent.com/17034772/122280192-3e213e80-cee9-11eb-8458-ab8caa5968d5.mp4

FTR: I think the widget shouldn't support deletion or any other kind of mutation and should be just a presentation layer. This way we force the user to document any relevant mutations as code in Elixir cells, which is in the spirit of reproducability.

### DIFF
--- a/lib/kino/ets.ex
+++ b/lib/kino/ets.ex
@@ -68,13 +68,19 @@ defmodule Kino.ETS do
 
   @impl true
   def handle_info({:connect, pid}, state) do
+    table_name = :ets.info(state.tid, :name)
+    name = "ETS #{inspect(table_name)}"
+
     columns =
       case :ets.match_object(state.tid, :_, 1) do
         {[record], _} -> columns_structure_from_record(record)
         :"$end_of_table" -> []
       end
 
-    send(pid, {:connect_reply, %{columns: columns, features: [:pagination, :sorting]}})
+    send(
+      pid,
+      {:connect_reply, %{name: name, columns: columns, features: [:pagination, :sorting]}}
+    )
 
     {:noreply, state}
   end
--- a/lib/kino/ets.ex
+++ b/lib/kino/ets.ex
@@ -1,0 +1,145 @@
+defmodule Kino.ETS do
+  @moduledoc """
+  A widget for interactively viewing an ETS table.
+
+  ## Examples
+
+      tid = :ets.new(:users, [:set, :public])
+      Kino.ETS.start(tid)
+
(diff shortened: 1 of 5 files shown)


josevalim:
This looks great!!! :heart: Should we implement Kino.Render for Reference and have it call :ets.info(...) on it so we automatically format ETS tables? I would personally enjoy that.

I also think the table should support a title. Perhaps ETS table or ETS #Reference<...>.

josevalim:
Oh, we should also support rendering a list of structs as a table, but that's probably a static table. :)

josevalim:
And another neat use case would be to support an ecto query and a repo! omg, i am very excited, haha!

jonatanklosko:
> Should we implement Kino.Render for Reference and have it call :ets.info(...) on it so we automatically format ETS tables?

Can't think of a good reason not to do that and it seems pretty cool, so let's have that!

> I also think the table should support a title. Perhaps ETS table or ETS #Reference<...>.

I like that, especially that in some cases there may be a clear name for the given table.

> Oh, we should also support rendering a list of structs/maps as a table, but that's probably a static table. :)

This probably depends on data size. If there are a lot of entries, then I think it would be better to have LV query for small row batches rather than sending all the rows as part of the output and duplicating whenever the notebook is sent between processes. So perhaps even that static case should be dynamic (dynamic, since we don't show all rows at once)?

> And another neat use case would be to support an ecto query and a repo!

Haha, this idea kept spamming my head working on the tables :joy_cat:

josevalim:
To be clear, the list of maps/structs should be opt-in.  But for :ets I think it is fine to always do that!

And now that I think about it... the lists of maps/structs is still a dynamic table, we likely don't want to send everything to the client at once.

jonatanklosko:
WDYT?

image

We need to start the widget process implicitly, but that will be still done in Evaluator, so monitoring works the same way. This seems fine since we ensure the given `tid` should work with the widget beforehand.

jonatanklosko:
Also, showing table name (we can get the actual name with `:ets.info` :D):

image

josevalim (approved):

PR #22, opened 2021-06-21
Implement DataTable

### LINKED ISSUE #17: Add a table function that works with any enums
As long as we have an enum of tuples or enum of maps, we can render them with pagination and what not.

For maps, field ordering is going to be sorted. For keywords, we respect the keyword ordering. For tuples, it is the indexes as done in ETS. And we can also accept options such as `:fields` when calling DataTable/data_table.

jonatanklosko:
Closes #17.

https://user-images.githubusercontent.com/17034772/122764304-a41e1500-d29f-11eb-8cc3-541b941c1ffa.mp4

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -1,0 +1,223 @@
+defmodule Kino.DataTable do
+  @moduledoc """
+  A widget for interactively viewing enumerable data.
+
+  The data must be an enumerable of records, where each
+  record is either map, keyword list or tuple. Structs
+  however must be manually converted to maps.
+
+  ## Examples
+
+      data = [
+        %{id: 1, name: "Elixir", website: "https://elixir-lang.org"},
+        %{id: 2, name: "Erlang", website: "https://www.erlang.org"}
+      ]
+
+      Kino.DataTable.start(data)
+
+  The tabular view allows you to quickly preview the data
+  and analyze it thanks to sorting capabilities.
+
+      data = Process.list() |> Enum.map(&Process.info/1)
+
+      Kino.DataTable.start(
+        data,
+        keys: [:registered_name, :initial_call, :reductions, :stack_size]
+      )
+  """
+
+  use GenServer, restart: :temporary
+
+  defstruct [:pid]
+
+  @type t :: %__MODULE__{pid: pid()}
+
+  @typedoc false
+  @type state :: %{
+          parent_monitor_ref: reference(),
+          data: Enum.t(),
+          total_rows: non_neg_integer()
+        }
+
+  @doc """
+  Starts a widget process with enumerable tabular data.
+  """
+  @spec start(Enum.t(), keyword()) :: t()
+  def start(data, opts \\ []) do
+    validate_data!(data)
+
+    parent = self()
+    keys = opts[:keys]
+    opts = [data: data, parent: parent, keys: keys]
+
+    {:ok, pid} = DynamicSupervisor.start_child(Kino.WidgetSupervisor, {__MODULE__, opts})
+
+    %__MODULE__{pid: pid}
+  end
+
+  defp validate_data!(data) do
+    Enum.reduce(data, nil, fn record, type ->
+      case record_type(record) do
+        :struct ->
+          raise ArgumentError,
+                "struct records are not supported, you need to convert them to maps explicitly"
+
+        :other ->
+          raise ArgumentError,
+                "expected record to be either map, tuple or keyword list, got: #{inspect(record)}"
+
+        first_type when type == nil ->
+          first_type
+
+        ^type ->
+          type
+
+        other_type ->
+          raise ArgumentError,
+                "expected records to have the same data type, found #{type} and #{other_type}"
+      end
+    end)
+  end
+
+  defp record_type(record) do
+    cond do
+      is_struct(record) -> :struct
+      is_map(record) -> :map
+      is_tuple(record) -> :tuple
+      Keyword.keyword?(record) -> :keyword_list
+      true -> :other
+    end
+  end
+
+  @doc false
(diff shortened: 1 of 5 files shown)


lib/kino/data_table.ex
```diff
@@ -0,0 +1,223 @@
...
+
+  @typedoc false
+  @type state :: %{
+          parent_monitor_ref: reference(),
+          data: Enum.t(),
+          total_rows: non_neg_integer()
+        }
+
+  @doc """
+  Starts a widget process with enumerable tabular data.
```
josevalim, line 43: We need to document the options!

jonatanklosko: Whoops, forgot in the end!

lib/kino/data_table.ex
```diff
@@ -0,0 +1,223 @@
...
+  """
+  @spec start(Enum.t(), keyword()) :: t()
+  def start(data, opts \\ []) do
+    validate_data!(data)
```
josevalim, line 47: I don't want to validate the data upfront because I want to support passing lazy enumerables. If we do a full traversal upfront, then that can be an issue. :)

josevalim: It is best to validate on every page, as we render them.

lib/kino/data_table.ex
```diff
@@ -0,0 +1,223 @@
...
+    parent = Keyword.fetch!(opts, :parent)
+    keys = Keyword.fetch!(opts, :keys)
+
+    parent_monitor_ref = Process.monitor(parent)
+
+    columns =
+      if keys do
+        Enum.map(keys, &key_to_column/1)
+      else
+        columns_structure(data)
```
josevalim, line 109: Same here, let's compute this per page and not on init.

jonatanklosko: We support sorting, so we do the traversal on every sort request anyway, don't we?

lib/kino/data_table.ex
jonatanklosko: This is the only reasonable place to validate though, before the process is started. Should we assume the data is valid and then let the widget crash silently?

lib/kino/data_table.ex
josevalim: Can we make sorting optional and enable it by default only if we have a list? And then we clarify that sorting requires traversing the whole table?

Btw, do we support sorting on ETS tables? If so, what happens if we sort on a column that does not exist on all rows? Do we show missing data first or last?

lib/kino/data_table.ex
josevalim: Perhaps we can validate if it is a list? And then skip it otherwise? You can add tests with a `MapSet` of maps as counter examples...

jonatanklosko: Sounds good to me!

lib/kino/data_table.ex
jonatanklosko: Optional sorting sounds perfect!

As for ETS sorting we do in fact fail. I used `:qlc.keysort` which seemed appropriate, but it does error if the position doesn't exist on an entry:

https://github.com/elixir-nx/kino/blob/ed0ae420301f75e058d60fc4c154178e0b5be980/lib/kino/ets.ex#L131

josevalim: @jonatanklosko hrm, should we address that somehow?

jonatanklosko: I cannot see any built-in sorting that would allow us to address this and we definitely don't want to read all the data into list and sort ourselves. This doesn't seem like a very likely case, so I think it's better to have the edge case failing rather than disabling sorting altogether.

Perhaps when retrieving rows we should catch the error and then do another retrieval without sorting?

josevalim: @jonatanklosko it may be worth looking into qlc because I am almost sure if it sorting the whole table...

josevalim: Perhaps the optimization we can do is to not keep everything in memory as we sort. Instead, we only need to keep limit+offset in memory. Perhaps it is a feature that could even be added to Elixir itself.

jonatanklosko: > I am almost sure if it sorting the whole table

We are not really limiting the rows anyway, so we need to sort all of them this way of another. From the docs it's not exactly clear to me how qlc sorts results in case of ETS

> [The sorter uses temporary files only if QH1 does not evaluate to a list and the size of the binary representation of the answers exceeds Size bytes, where Size is the value of option size](https://erlang.org/doc/man/qlc.html#keysort-2)

Specifically, I'm not sure how to interpret "if the query handler QH1 evaluates to a list", if that's the case for ETS or not.

josevalim: I see. I would just make sure it works for now and then refactor later.

jonatanklosko: My point is that the only way I see is to convert the table to list and do regular sorting, but that doesn't sound like a good idea ^^

(This PR should be ready btw)

josevalim: Right, so I would either disable sorting or convert to a list.

for this case though, how does sorting nils behave? Do they come first or last?

jonatanklosko: > Right, so I would either disable sorting or convert to a list.

You mean it's alright if we `:ets.tab2list` and then do usual sorting and slicing? What about a really huge table?

> for this case though, how does sorting nils behave? Do they come first or last?

The fields can be any terms, they are inspected only for displaying, so they are sorted according to the usual rules, meaning `nil` is first, except when we deal with numbers.

josevalim: Tab2list is alright for now but ideally for sorting we would do a take_sort.

josevalim (approved):

PR #26, opened 2021-06-24
Make inspect options globally configurable

jonatanklosko:
See https://github.com/elixir-nx/livebook/issues/392#issuecomment-867590608.

image

### DIFF
--- a/lib/kino/config.ex
+++ b/lib/kino/config.ex
@@ -8,6 +8,11 @@ defmodule Kino.Config do
     Application.get_all_env(:kino) |> Keyword.take(@keys)
   end
 
+  @spec configuration(atom()) :: term()
+  def configuration(key, default \\ nil) do
+    Application.get_env(:kino, key, default)
+  end
+
   @spec configure(keyword()) :: :ok
   def configure(options) do
     Enum.each(options, &validate_option/1)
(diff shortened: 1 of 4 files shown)


lib/kino/output.ex
```diff
@@ -177,9 +177,10 @@ defmodule Kino.Output do
     text_block(inspected)
   end
 
-  defp inspect_opts(opts \\ []) do
+  defp inspect_opts() do
     default_opts = [pretty: true, width: 100, syntax_colors: syntax_colors()]
-    Keyword.merge(default_opts, opts)
+    config_opts = Kino.Config.configuration() |> Keyword.get(:inspect, [])
```
josevalim, line 182: We should probably go with:

```suggestion
    config_opts = Kino.Config.configuration(:inspect, [])
```

There is no reason to read the whole Kino configuration table for a single key. :)

lib/kino/config.ex
```diff
@@ -8,6 +8,11 @@ defmodule Kino.Config do
     Application.get_all_env(:kino) |> Keyword.take(@keys)
   end
 
+  @spec configuration(atom()) :: term()
+  def configuration(key, default \\ nil) do
```
josevalim, line 12: If you want to be double sure...

```suggestion
  def configuration(key, default \\ nil) when key in @keys do
```

But optional. :)

josevalim (approved):

PR #33, opened 2021-07-21
Support structs in DataTable

### LINKED ISSUE #32: Support structs in data_table
We should not show fields starting with underscore by default. We could add a `show_underscored: true` if someone wants to disable this behaviour. This will allow us to cleanly handle things like `__struct__` and `__meta__` fields.

jonatanklosko:
Closes #32.

image

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -51,6 +51,9 @@ defmodule Kino.DataTable do
       Sorting requires traversal of the whole enumerable, so it may not be
       desirable for lazy enumerables. Defaults to `true` if data is a list
       and `false` otherwise.
+
+    * `:show_underscored` - whether to include record keys starting with underscore.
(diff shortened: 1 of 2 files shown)


lib/kino/data_table.ex
```diff
@@ -212,7 +226,7 @@ defmodule Kino.DataTable do
     |> Enum.map(fn {_, idx} -> key_to_column(idx) end)
   end
 
-  defp columns_structure_for_record(record) when is_map(record) do
+  defp columns_structure_for_record(record) when is_map(record) or is_struct(record) do
```
josevalim, line 229: `is_map(record)` will match all structs already:

```suggestion
  defp columns_structure_for_record(record) when is_map(record) do
```

josevalim (commented):
Two minor comments and ship it!

PR #34, opened 2021-07-22
Implement table widget for Ecto queries

jonatanklosko:
Mostly useful with Kino as a dev dependency in Mix projects, but may also be useful for Ecto tutorial notebooks.

https://user-images.githubusercontent.com/17034772/126636714-45b7495a-31a0-4400-9242-d7f60b9ed81a.mp4


jonatanklosko:
[Here](https://gist.github.com/jonatanklosko/5dfedc5bfd88c3143ce01a19192b574e) is a self-contained example notebook.

lib/kino/ecto.ex
```diff
@@ -0,0 +1,152 @@
...
+        query
+      end
+
+    records = repo.all(query)
+
+    {count, records}
+  end
+
+  defp default_select_query?(queryable) do
+    query = from(q in queryable, [])
```
josevalim, line 149: Call to_queryable instead?

lib/kino/utils/table.ex
```diff
@@ -0,0 +1,121 @@
...
+
+    # Note: id is opaque to the client, and we don't need it for now
+    %{id: nil, fields: fields}
+  end
+
+  @doc """
+  Checks if the given term is an Ecto.Schema.
+  """
+  def ecto_schema?(queryable) do
+    is_atom(queryable) and function_exported?(queryable, :__schema__, 1)
```
josevalim, line 119: Maybe it is better to convert to a queryable and then look under the from?

lib/kino/utils/table.ex
```diff
@@ -0,0 +1,121 @@
...
+      end)
+
+    # Note: id is opaque to the client, and we don't need it for now
+    %{id: nil, fields: fields}
+  end
+
+  @doc """
+  Checks if the given term is an Ecto.Schema.
+  """
+  def ecto_schema?(queryable) do
```
josevalim, line 118: Is this reused anywhere? If not, maybe we should move to the Ecto module?

lib/kino/utils/table.ex
```diff
@@ -0,0 +1,121 @@
+defmodule Kino.Utils.Table do
```
josevalim, line 1: Maybe Kino.Table.Utils? It feels Table is more important but that’s a nitpick. :)

lib/kino/utils/table.ex
jonatanklosko: I may be missing something, but the information doesn't seem to be explicitly there:

image

lib/kino/utils/table.ex
jonatanklosko: It's used in this module in `columns_for_record`.

lib/kino/ecto.ex
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  def handle_info({:DOWN, ref, :process, _object, _reason}, %{parent_monitor_ref: ref} = state) do
+    {:stop, :shutdown, state}
+  end
+
+  defp get_records(repo, queryable, rows_spec) do
+    count = repo.aggregate(queryable, :count)
+
+    query = from(q in queryable, limit: ^rows_spec.limit, offset: ^rows_spec.offset)
```
josevalim, line 133: Should we check the query has no limit and offset upfront? Otherwise this will raise. Alternatively we can call Query.exclude and remove both values upfront.

lib/kino/ecto.ex
```diff
@@ -0,0 +1,152 @@
...
+        {"Query", []}
+      end
+
+    features =
+      Kino.Utils.truthy_keys(
+        refetch: true,
+        pagination: true,
+        # If the user specifies custom select, the record keys
+        # are not valid "order by" fields, so we disable sorting
+        sorting: default_select_query?(state.queryable)
```
josevalim, line 95: Should we sort if there is already sorting?

lib/kino/ecto.ex
jonatanklosko: Does it raise thought? It looks like the values are just overridden:

image

lib/kino/ecto.ex
jonatanklosko: I should be fine. We show the initially sorted results, but the user can also click on columns to override the sorting.

lib/kino/utils/table.ex
jonatanklosko: My reasoning is that `Utils` is less specific and can thus serve as a namespace, no strong opinion though.

josevalim: Fine by me :)

lib/kino/ecto.ex
josevalim: The issue is that we are not overriding, we are adding to the sorting. We would need to exclude the existing sorting first. We should have tests for this :)

jonatanklosko: Oh right! Exclude sounds good. Is there some test Repo that we could use or how would you go about the tests?

josevalim (commented):
Comments added. We should probably devise a testing scheme where we emulate the messages to handle some of the corner cases.

lib/kino/ecto.ex
josevalim: There is an ETSO adapter for ETS, no dependencies but I am not sure if it supports order by. Otherwise either PG or mocks.

lib/kino/ecto.ex
josevalim: Ah, I thought we raised. Good then!

lib/kino/utils/table.ex
josevalim: Hrm, not sure if it is worth having it there. It is like the handling of Ecto queries is now everywhere instead of isolated in a single place. Consider splitting that apart or having Ecto not call it?

Also remember that Ecto’s select can return a tuple, so we need to take that into account too.

jonatanklosko: This was kinda intentional, because if someone passes a list of Ecto structs into `DataTable`, we may as well use the fields information it provides, right?

> Also remember that Ecto’s select can return a tuple, so we need to take that into account too.

Yeah, it can be a map or a single value too, that's the reason why I moved much stuff to `Table.Utils`, because it's pretty much the same for `DataTable` and `ETS`.

josevalim: > This was kinda intentional, because if someone passes a list of Ecto structs into DataTable, we may as well use the fields information it provides, right?

i would just treat them as structs since the fields are in the struct too? What benefits would be added by looking into Ecto?

jonatanklosko: The benefits are that the columns are ordered as per the schema definition, which should feel more natural. Also this way we don't show virtual fields such as associations.

josevalim: Hrm, I see. Sounds good to me then!

lib/kino/utils/table.ex
jonatanklosko: Ah, as far as I can see the second element in source is either a schema or `nil`, so this helps. I updated the extraction in ca7064448bf3f42f05374a15484f6c4cee52852c.

lib/kino/utils/table.ex
```diff
@@ -113,9 +113,27 @@ defmodule Kino.Utils.Table do
...
-  def ecto_schema?(queryable) do
-    is_atom(queryable) and function_exported?(queryable, :__schema__, 1)
+  def ecto_schema(queryable)
+
+  def ecto_schema(struct) when is_struct(struct) do
+    ecto_schema(struct.__struct__)
+  end
+
+  def ecto_schema(queryable) when is_atom(queryable) do
+    if function_exported?(queryable, :__schema__, 1) do
```
josevalim, line 127:
```suggestion
    if Code.ensure_loaded?(queryable) and function_exported?(queryable, :__schema__, 1) do
```

lib/kino/utils/table.ex
```diff
@@ -0,0 +1,139 @@
...
+      queryable
+    else
+      nil
+    end
+  end
```
jonatanklosko, line 132: We could use `Ecto.Queryable.to_query`, but we don't want this function to rely on Ecto being installed.

lib/kino/ecto.ex
jonatanklosko: There's an old [pending PR](https://github.com/evadne/etso/pull/12) that adds ORDER BY support ot ETSO. Either way, to make precise assertions we need mocking in some form. I've just introduced a `MockRepo` that we can assert and resolve calls against, let me know if it makes sense :)

josevalim (approved):

PR #42, opened 2021-08-19
Add Kino.Explorer

jonatanklosko:
We may want to wait until the official release, but doesn't hurt to have the implementation around :)

https://user-images.githubusercontent.com/17034772/130112186-b4ebcea6-821a-4623-ac44-62fe66384e6b.mp4


lib/kino/explorer.ex
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  def handle_info({:DOWN, ref, :process, _object, _reason}, %{parent_monitor_ref: ref} = state) do
+    {:stop, :shutdown, state}
+  end
+
+  defp get_records(df, rows_spec) do
+    df =
+      if order_by = rows_spec[:order_by] do
+        Explorer.DataFrame.arrange(df, [{order_by, rows_spec.order}])
```
jonatanklosko, line 89: In particular, this will change to `[{rows_spec.order, order_by}]` soon as far as I understand.

josevalim (commented):
LGTM! But we can’t merge it yet because the Git dep wouldn’t let us publish new versions. So I would wait.

We also need to decide if this is the default behavior or not. :) the consensus was that it should be, but in this case, we need changes to LV.

josevalim:
From our slack discussion: let's show the string representation for values and column names.

jonatanklosko:
With a tiny change to Livebook (https://github.com/livebook-dev/livebook/commit/76f84eb66d11f7ff38aa7d92265a9515d7a5c76e):

image

jonatanklosko:
cc @cigrainger

cigrainger:
This is _very cool_! I love the dtypes representation.

lib/kino/explorer.ex
cigrainger: That's right! `order_by` is a `List` in either case.

cigrainger:
This closes https://github.com/elixir-nx/explorer/issues/21, right?

lib/kino/explorer.ex
cigrainger: I'm actually going to more closely match the `Ecto.Query` behaviour as José suggested, so it won't be a list. PR to follow.

jonatanklosko: IIRC the point was to replace `name: :asc, age: :desc` by `asc: "name", desc: "age"`, so it's still a keyword list? For Ecto we do:

https://github.com/livebook-dev/kino/blob/c38ee8750361cc0a586b67049a5432db0fd1dfe0/lib/kino/ecto.ex#L157-L158

cigrainger: That's right. I was saying it could be a keyword list with the value as a list so: `Keyword.t([String.t()])`, e.g. `asc: ["name1", "name2"], desc: ["age"]` but went with the way you described it above. See elixir-nx/explorer#39.

jonatanklosko: Great! Updated here accordingly :)

lib/kino.ex
```diff
@@ -90,6 +90,14 @@ defmodule Kino do
 
       Kino.Ecto.new(Weather, Repo)
 
+  ### Kino.Explorer
+
+  `Kino.Explorer` implements a data table output for [Explorer](https://github.com/amplifiedai/explorer)
```
ankopainting, line 95: should this be https://github.com/elixir-nx/explorer ?

jonatanklosko:
@ankopainting it's fixed on Livebook main (https://github.com/livebook-dev/livebook/pull/991) :)

PR #49, opened 2021-11-08
Add Kino.Frame for animations

### LINKED ISSUE #48: Add Dynamic Image widget
Hi Dashbit friends.

I am experimenting with a dynamic image widget and wondering whether it'd make sense to submit a pull request to have this be part of Kino / Livebook.

## Rationale

Initially I needed it show a video feed in one of my nerves notebooks with:

```elixir
video = fn y ->
  # Picam  captures stills using the camera module on Raspberry Pi
  Kino.Image.new(Picam.next_frame, :jpeg)
  |> Kino.render

  Process.sleep(100)
  y.(y)
end

video.(video)
```

My intention was to take a photo every 100ms, then update an image output cell with that image, but rendering `Kino.Image` appends a new image to the notebook instead.

There are though numerous use-cases, especially educational, where a dynamic image can be used with an `svg` to visualise an algorithm in an animated fashion. I'm am already working on visualising towers of hanoi and game of life.

## Demo

https://user-images.githubusercontent.com/645514/140661823-93b93bc0-ee0e-4389-9d07-0b3b6616a1a7.mp4

## Implementation

Currently, my implementation consists of a `LivebookWeb.Output.ImageLive` view in `livebook` and a `Kino.ImageDynamic` GenServer and minor changes in a few other files.

Please let me know your thoughts and thank you for Livebook and Kino 🙌

jonatanklosko:
Closes #48. For the Livebook counterpart see https://github.com/livebook-dev/livebook/pull/688.

This introduces a new dynamic output that wraps a single static output, which can be replaced at any time. Additionally, we now have `Kino.animate` wrapping the most common use case.

https://user-images.githubusercontent.com/17034772/140779100-c21ebb5e-422e-41cd-acb0-42566bcd6da1.mp4

*Sidenote: `Kino.animate` is how to flip book in Livebook*

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -152,4 +152,48 @@ defmodule Kino do
   def configure(options) do
     Kino.Config.configure(options)
   end
+
+  @doc ~S"""
+  Returns a widget that periodically calls the given function
+  to render a new result.
+
+  The callback is run every `interval_ms` milliseconds and receives
+  the accumulated value. The callback should return either of:
+
+    * `{:cont, term_to_render, acc}` - the continue with the new
+      accumulated value
+
+    * `:halt` - to no longer schedule callback evaluation
+
+  This function uses `Kino.Frame` as the underlying widget.
+
+  ## Examples
+
+      # Render new Markdown every 100ms
+      Kino.animate(100, 0, fn i ->
+        md = Kino.Markdown.new("**Iteration: `#{i}`**")
+        {:cont, md, i + 1}
+      end)
+  """
+  @spec animate(
+          pos_integer(),
+          term(),
+          (term() -> {:cont, term(), acc :: term()} | :halt)
+        ) :: Kino.Frame.t()
+  def animate(interval_ms, acc, fun) do
+    widget = Kino.Frame.new()
+
+    Kino.Frame.periodically(widget, interval_ms, acc, fn acc ->
+      case fun.(acc) do
+        {:cont, term, acc} ->
+          Kino.Frame.render(widget, term)
+          {:cont, acc}
+
+        :halt ->
+          :halt
+      end
+    end)
+
+    widget
+  end
 end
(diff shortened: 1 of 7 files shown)


lib/kino.ex
```diff
@@ -152,4 +152,48 @@ defmodule Kino do
...
   end
+
+  @doc ~S"""
+  Returns a widget that periodically calls the given function
+  to render a new result.
+
+  The callback is run every `interval_ms` milliseconds and receives
+  the accumulated value. The callback should return either of:
+
+    * `{:cont, term_to_render, acc}` - the continue with the new
```
zorbash, line 163: `to continue`

josevalim:
I love the name and the abstraction!

Some concerns:

1. Is it too much to have both the public module and `Kino.animate`? Shall we `@moduledoc false` this one? Or do we need to keep it public because the result of `Kino.animate` is public?
2. `Kino.render` does not return an output. `Kino.VegaLite.periodically` only returns ok. The fact animate does is a bit inconsistent?
3. Should we rename `Kino.VegaLite.periodically` to `Kino.VegaLite.animate` for consistency?

jonatanklosko:
@josevalim

1. We still want `Kino.Frame` to be public, because someone may use it explicitly, specifically the `Kino.Frame.render` function. This can be useful if the rendering is not periodical, but rather based on some event, just to give an example, someone may render some information from an Axon callback.
2. Ah, initially I thought we want to return the widget to render it, but you're absolutely right, we should call `Kino.render` from `Kino.animate`, so it works anywhere in the cell.
3. For the widgets, `periodically` is actually more accurate. The callback we pass to `periodically` doesn't any new data to animate, it needs to be done using `Kino.VegaLite.{push,push_many}` or `Kino.Frame.render`.

josevalim (approved):

PR #50, opened 2021-11-19
Add Kino.Controls for user interactions

jonatanklosko:
For the Livebook counterpart see https://github.com/livebook-dev/livebook/pull/710.

Introduces UI controls that allow the user to feed events into any process that subscribes. I built a multiplayer Pong as a driving showcase, but we already have some cooler things in mind :smile_cat:

**Basic usage**

https://user-images.githubusercontent.com/17034772/142630044-f16867de-3772-4b6d-ad9c-fcdf426336f6.mp4

**Notebook preview**

https://user-images.githubusercontent.com/17034772/142630078-4b0eead5-2919-411e-acd5-f3a8aa40eb9a.mp4

**The game**

https://user-images.githubusercontent.com/17034772/142630120-ed86c652-87a5-40d9-bc96-863a62de1915.mp4


jonatanklosko:
[Here](https://gist.github.com/jonatanklosko/11218655df00f8472e45d4f6e973743e) is the current notebook. I think that's something we could have in the Explore section.

josevalim:
This is absolutely amazing!

My only comment for now is, instead of having a control with different kinds of inputs, we should have different controls and find a mechanism for rendering multiple elements in the same frame. Perhaps it could be as simple as allowing a list in frame? Some elements could be inline-block by default and others block, which means they also get nicely formatted.

jonatanklosko:
I split it into individual controls and implemented subscription for controls and inputs. There is still a number of things, like anonymous functions and un-sync, but I've just pushed the changes, so let me know what you think of the API.

In particular I made it so that the event name is specified when subscribing, like `Kino.Control.subscribe(button, :hello)`, while previously it would be a part of the button. The primary reason is that otherwise we would need to pass event name when creating inputs for consistency, which feels backwards and in many cases unnecessary.

https://user-images.githubusercontent.com/17034772/143957419-73890e2a-fcfd-4ee1-a60f-72808565f9f6.mp4

lib/kino/subscription_manager.ex
```diff
@@ -0,0 +1,95 @@
...
+    GenServer.cast(@name, {:unsubscribe, topic, pid})
+  end
+
+  @impl true
+  def init(_opts) do
+    {:ok, %{subscribers_by_topic: %{}}}
+  end
+
+  @impl true
+  def handle_cast({:subscribe, topic, pid, receive_as}, state) do
```
jonatanklosko, line 51: Here `topic` is the input/control id that we subscribe to. We monitor the subscriber and remove it accordingly when it goes down, however we don't detect when the input/control (topic) no longer exists.

With processes it may not be _as_ critical, since usually we would expect the subscribing process to die on reevaluation, but once we start supporting functions we would keep those forever.

So it would be nice to detect when the inputs are removed and one solution I see is for the session to look at the outputs it removes and if there are any controls/inputs it sends a termination event to `destination`, just we do with regular events. WDYT?

lib/kino/control.ex
```diff
@@ -0,0 +1,126 @@
...
+
+  @type t :: %__MODULE__{attrs: Kino.Output.control_attrs()}
+
+  defp new(attrs) do
+    token = Kino.Bridge.generate_token()
+    id = {token, attrs} |> :erlang.phash2() |> Integer.to_string()
+
+    attrs =
+      Map.merge(attrs, %{
+        id: id,
```
jonatanklosko, line 43: We should actually have an unique id for subscriptions, I will adjust once the garbage collection is in place as per the other comment.

And for inputs we will need both `id` and reevaluation-safe-id (maybe `key` or `hash` :see_no_evil:).

lib/kino/control.ex
```diff
@@ -0,0 +1,126 @@
...
+  either by placing it at the end of a code cell or by
+  explicitly rendering it with `Kino.render/1`.
+
+      button = Kino.Control.button("Hello")
+
+  Next, to receive events from the control, a process needs to
+  subscribe to it and specify pick a name to distinguish the
+  events.
+
+      Kino.Control.subscribe(button, :hello)
```
josevalim, line 23: What if instead of giving a name, `subscribe` returns a reference. And you can use this reference to associate to a certain client/name?

lib/kino/control.ex
```diff
@@ -0,0 +1,126 @@
...
+      #=> {:event, :hello, %{origin: #PID<10895.9854.0>}}
+  """
+
+  defstruct [:attrs]
+
+  @type t :: %__MODULE__{attrs: Kino.Output.control_attrs()}
+
+  defp new(attrs) do
+    token = Kino.Bridge.generate_token()
+    id = {token, attrs} |> :erlang.phash2() |> Integer.to_string()
```
josevalim, line 39:
```suggestion
    id = make_ref()
```

lib/kino/control.ex
josevalim, line 23: Let's call it tag instead of name.

lib/kino/control.ex
josevalim: Sounds good to me!

lib/kino.ex
```diff
@@ -193,4 +212,20 @@ defmodule Kino do
...
+  Ties the given process lifetime to the caller.
+
+  When used directly in a Livebook cell, the process is killed
+  on re-evaluation.
+
+  When used from another process, the given process is killed
+  as soon as the parent terminates.
+  """
+  @spec bind_process(pid()) :: :ok | {:error, atom()}
+  def bind_process(pid) do
```
jonatanklosko, line 226: Not sure if we want to expose this directly or only as `Kino.start_supervised` or `Kino.start_child` as in #51.

lib/kino/control.ex
```diff
@@ -0,0 +1,137 @@
...
+
+    * `:type` - either `:keyup` or `:keydown`
+
+    * `:key` - the value matching the browser [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)
+
+  ### Status event
+
+    * `:type` - either `:status`
+
+    * `:enabled` - whether the keyboard is activated
```
jonatanklosko, line 81: Not sure if it's better to have two separate events or a status event :thinking:

lib/kino/control.ex
```diff
@@ -0,0 +1,137 @@
...
+
+      Kino.Control.subscribe(keyboard, :keyboard)
+
+  As the user types events are streamed:
+
+      IEx.Helpers.flush()
+      #=> {:event, :keyboard, %{key: "o", origin: #PID<10895.9854.0>, type: :keydown}}
+      #=> {:event, :keyboard, %{key: "k", origin: #PID<10895.9854.0>, type: :keydown}}
+      #=> {:event, :keyboard, %{key: "o", origin: #PID<10895.9854.0>, type: :keyup}}
+      #=> {:event, :keyboard, %{key: "k", origin: #PID<10895.9854.0>, type: :keyup}}
```
josevalim, line 99: Remove `:event` prefix.

jonatanklosko:
I've updated the [Pong notebook](https://gist.github.com/jonatanklosko/11218655df00f8472e45d4f6e973743e) to match the new design.

lib/kino.ex
```diff
@@ -193,4 +212,60 @@ defmodule Kino do
...
+  """
+  @spec start_child(
+          Supervisor.child_spec()
+          | {module(), term()}
+          | module()
+        ) :: DynamicSupervisor.on_start_child()
+  def start_child(child_spec) do
+    # Starting a process that calls Kino.start_child/1 in its init
+    # would block forever, so we don't allow nesting
+    if Kino.DynamicSupervisor in Process.get(:"$ancestors", []) do
```
josevalim, line 230: Let's add a test for this to make sure it is triggered. Something like this should trigger it:

    Kino.start_child({Task, fn -> Kino.start_child({Task, fn -> :ok end}) end})

lib/kino.ex
```diff
@@ -193,4 +212,60 @@ defmodule Kino do
...
+    child_spec = %{child_spec | start: {Kino, :__start_override__, [start, parent]}}
+    DynamicSupervisor.start_child(Kino.DynamicSupervisor, child_spec)
+  end
+
+  @doc false
+  def __start_override__({mod, fun, args}, parent) do
+    # We switch the group leader, so that the newly started
+    # process gets the same group leder as the caller
+    initial_gl = Process.group_leader()
+    gl = Process.info(parent)[:group_leader]
```
josevalim, line 247: Pass the group leader instead of reading it through info. This has two issues:

1. This requires a lock in the parent process
2. This is going to copy a lot of information from the parent process (but this particular problem could be addressed by using Process.info/2

lib/kino.ex
```diff
@@ -193,4 +212,60 @@ defmodule Kino do
...
+      apply(mod, fun, args)
+      |> case do
+        {:ok, pid} = resp -> {:ok, resp, pid}
+        {:ok, pid, _info} = resp -> {:ok, resp, pid}
+        resp -> {:error, resp}
+      end
+      |> case do
+        {:ok, resp, pid} ->
+          Kino.Bridge.object_add_pointer(pid, parent)
+          Kino.Bridge.object_monitor(pid, Kino.Terminator.cross_node_name(), {:terminate, pid})
```
josevalim, line 261: Can we do this operation in a single pass?

Have object_monitor receive pid, parent, target and message. If there is no pointer, then you add one behind the scenes. The goal is to reduce the communication surface with Livebook.

```suggestion
          Kino.Bridge.object_add_pointer(pid, parent)
          Kino.Bridge.object_monitor(pid, Kino.Terminator.cross_node_name(), {:terminate, pid})
```

lib/kino.ex
```diff
@@ -193,4 +212,60 @@ defmodule Kino do
...
+
+    Process.group_leader(self(), gl)
+
+    try do
+      apply(mod, fun, args)
+      |> case do
+        {:ok, pid} = resp -> {:ok, resp, pid}
+        {:ok, pid, _info} = resp -> {:ok, resp, pid}
+        resp -> {:error, resp}
+      end
```
josevalim, line 257: Nitpick, return this:

```suggestion
      |> case do
        {:ok, pid} = resp -> {resp, pid}
        {:ok, pid, _info} = resp -> {resp, pid}
        resp -> {resp, nil}
      end
```

Then below you do:

```elixir
    {resp, pid} ->
      pid && Kino.Bridge.object_monitor(pid, parent, Kino.Terminator.cross_node_name(), {:terminate, pid})
      resp
```

lib/kino.ex
jonatanklosko: > Can we do this operation in a single pass?

Sure, but then we will essentially have `object_add_pointer_and_monitor` + `object_add_pointer` that we need in other places. So we would optimise this case into a single message, but `object_monitor` is really a separate action that we could use for something else in the future.

> The goal is to reduce the communication surface with Livebook.

The surface doesn't change though, since we still need at least two actions?

> Also, we only monitor PIDs right? So maybe we should remove object_ from the naming.

This can be anything, in particular we monitor control/input refs to notify SubscriptionManager once it disappears.

lib/kino.ex
jonatanklosko: Good call! We need the assertion in the inner call, so I made the test wait for the outer process to terminate, let me know if there's any better way :)

lib/kino/subscription_manager.ex
```diff
@@ -0,0 +1,100 @@
...
+    {_, state} = pop_in(state.subscribers_by_topic[topic])
+    {:noreply, state}
+  end
+
+  def handle_info({:DOWN, _ref, :process, pid, _reason}, state) do
+    subscribers_by_topic =
+      state.subscribers_by_topic
+      |> Enum.map(fn {topic, subscribers} -> {topic, remove_pid(subscribers, pid)} end)
+      |> Enum.filter(&match?({_, []}, &1))
+      |> Map.new()
```
josevalim, line 92: Something we usually do, to avoid traversing all topics, is to have two maps: one %{topic => [pid]} and another %{pid => [topic]}, so you can always easily find all topics used by a certain pid.

josevalim (approved):
I just dropped one comment, plus the changes we discussed in Slack, then we are good to go!

lib/kino/subscription_manager.ex
jonatanklosko: Good idea, done!

PR #53, opened 2021-11-25
Introduce Kino.Input

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/714.

Currently `Kino.Input.read` raises in the terminal, but we can easily improve it to read and parse input. This can be handled separately though.

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -1,0 +1,228 @@
+defmodule Kino.Input do
+  @moduledoc """
+  Various input elements for entering data.
+
+  ## Examples
+
+  First, create an input and make sure it is rendered,
+  either by placing it at the end of a code cell or by
+  explicitly rendering it with `Kino.render/1`.
+
+      input = Kino.Input.text("Name")
+
+  Then read the value at any later point:
+
+      name = Kino.Input.read(input)
+  """
+
+  defstruct [:attrs]
+
+  @type t :: %__MODULE__{attrs: map()}
+
+  defp new(attrs) do
+    token = Kino.Bridge.generate_token()
+    id = {token, attrs} |> :erlang.phash2() |> Integer.to_string()
+    attrs = Map.put(attrs, :id, id)
+    %__MODULE__{attrs: attrs}
+  end
+
+  @doc """
+  Creates a new text input.
+
+  ## Options
+
+    * `:default` - the initial input value. Defaults to `""`
+
+    * `:multiline` - whether the input should be multiline
+      text box. Defaults to `false`
+  """
+  @spec text(String.t(), keyword()) :: t()
+  def text(label, opts \\ []) when is_binary(label) and is_list(opts) do
+    default = Keyword.get(opts, :default, "")
+    multiline = Keyword.get(opts, :multiline, false)
+    new(%{type: :text, label: label, default: default, multiline: multiline})
+  end
+
+  @doc """
+  Creates a new password input.
+
+  This is similar to text input, except the content is not
+  visible by default.
+
+  ## Options
+
+    * `:default` - the initial input value. Defaults to `""`
+  """
+  @spec password(String.t(), keyword()) :: t()
+  def password(label, opts \\ []) do
+    default = Keyword.get(opts, :default, "")
+    new(%{type: :password, label: label, default: default})
+  end
+
+  @doc """
+  Creates a new number input.
+
+  The input value is can be either a float or `nil`.
+
+  ## Options
+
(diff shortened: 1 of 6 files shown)


lib/kino/input.ex
```diff
@@ -0,0 +1,228 @@
...
+  @spec password(String.t(), keyword()) :: t()
+  def password(label, opts \\ []) do
+    default = Keyword.get(opts, :default, "")
+    new(%{type: :password, label: label, default: default})
+  end
+
+  @doc """
+  Creates a new number input.
+
+  The input value is can be either a float or `nil`.
```
josevalim, line 65: Is it always a float? 🤔

lib/kino/input.ex
```diff
@@ -0,0 +1,228 @@
...
+
+  This is similar to text input, except the content is not
+  visible by default.
+
+  ## Options
+
+    * `:default` - the initial input value. Defaults to `""`
+  """
+  @spec password(String.t(), keyword()) :: t()
+  def password(label, opts \\ []) do
```
josevalim, line 57: Guards missing!

lib/kino/input.ex
```diff
@@ -0,0 +1,228 @@
...
+      Kino.Input.select("Language", [{1, "One"}, {2, "Two"}, {3, "Three"}])
+  """
+  @spec select(String.t(), list({value :: term(), label :: String.t()}), keyword()) :: t()
+  def select(label, options, opts \\ [])
+      when is_binary(label) and is_list(options) and is_list(opts) do
+    if options == [] do
+      raise ArgumentError, "expected at least on option, got: []"
+    end
+
+    values = Enum.map(options, &elem(&1, 0))
```
josevalim, line 117: We should probably normalize all labels and values to strings here.

lib/kino/output.ex
```diff
@@ -149,6 +150,81 @@ defmodule Kino.Output do
...
+              type: :password,
+              id: input_id(),
+              label: String.t(),
+              default: String.t()
+            }
+          | %{
+              type: :number,
+              id: input_id(),
+              label: String.t() | nil,
+              default: number()
```
josevalim, line 191:
```suggestion
              label: String.t(),
              default: number() | nil
```

josevalim:
Looks great!

> Currently Kino.Input.read raises in the terminal, but we can easily improve it to read and parse input. This can be handled separately though.

Yeah, let's open up an issue!

lib/kino/input.ex
jonatanklosko: Yeah, it's a single number field so we always parse into float. Maybe we should have an option like `integer: true`?

lib/kino/input.ex
jonatanklosko: Labels yeah, but values can be anything no?

lib/kino/input.ex
josevalim: My understanding is actually the opposite, the number input is problematic on many browsers as not all of them support the decimal placeholder. We could try to read it as an integer and fallback to float only if the dot is there?

lib/kino/input.ex
josevalim: The problem is that the browser is going to send those as strings and doing the mapping will be annoying. Alternatively we support rendering only lists of values, such that: `select("Category", ~w(Books Cars ...))` and avoid this problem for now. Over the wire we still send it as a list of tuples though.

jonatanklosko: On Livebook side we use indices for the select and then map to the corresponding key, see [here](https://github.com/livebook-dev/livebook/blob/041764096c97414bc896a5067a3a5cd3901cb4ca/lib/livebook_web/live/output/input_component.ex#L51-L55).

lib/kino/input.ex
jonatanklosko: My point was rather than if we get `"10"`, `"10.0"`, `"10.1"` we just normalize all to float, so the type is consistent. But we can definitely try parsing as integer then as float. I was thinking we could have an option that makes the data type clear, but then we can also leave any conversion up to the user?

lib/kino/input.ex
josevalim: Ah, smart! So ignore me :)

lib/kino/input.ex
josevalim: > we can also leave any conversion up to the user?

Yeah, that's my current thinking right now. They may also want to use Decimal, for example.

josevalim (approved):

PR #60, opened 2021-12-05
Return the rendered term from Kino.render/1

### LINKED ISSUE #59: Add Kino.inspect/2 and return input on Kino.render/1
Kino.inspect/2 should behave just like IO.inspect, but with color.

jonatanklosko:
Closes #59.

@josevalim actually not sure if `Kino.inspect/2` makes sense with this change, because `Kino.render/1` covers most cases and if someone really wants to inspect Kino structs they can always use `IO.inspect` for that?

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -140,10 +140,14 @@ defmodule Kino do
   always produces colored text and respects the configuration
   set with `configure/1`.
   """
-  @spec inscpect(term()) :: term()
-  def inscpect(term) do
-    output = Kino.Output.inspect(term)
+  @spec inscpect(term(), keyword()) :: term()
+  def inscpect(term, opts \\ []) do
+    label = if label = opts[:label], do: "#{label}: ", else: ""
+
+    {:text, text} = Kino.Output.inspect(term, opts)
+    output = {:text, label <> text}
     Kino.Bridge.put_output(output)
+
     term
   end
 
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -28,7 +28,7 @@ defmodule Kino do
         |> Vl.data_from_series(...)
         |> ...
         |> Kino.VegaLite.new()
-        |> tap(&Kino.render/1)
+        |> Kino.render()
 
       Kino.VegaLite.push(widget, %{x: 1, y: 2})
 
@@ -95,7 +95,7 @@ defmodule Kino do
   `Kino.Frame` is a placeholder for static outptus that can
   be dynamically updated.
 
-      widget = Kino.Frame.new() |> tap(&Kino.render/1)
+      widget = Kino.Frame.new() |> Kino.render()
 
       for i <- 1..100 do
         Kino.Frame.render(widget, i)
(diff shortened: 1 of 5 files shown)


josevalim:
The goal of Kino.inspect is to be IO.inspect but with syntax coloring, so I would say it would be a welcome addition too. :)

jonatanklosko:
Yeah, my point is that `Kino.render/1` already does that most of the time. But I guess this will change as we start implementing the `Kino.Render` protocol for more library structures.

jonatanklosko:
For the record:

image
image

lib/kino.ex
```diff
@@ -116,20 +116,35 @@ defmodule Kino do
...
+
+  @doc """
+  Inspects the given term as cell output.
+
+  This works essentially the same as `IO.inspect/2`, except it
+  always produces colored text and respects the configuration
+  set with `configure/1`.
+  """
+  @spec inscpect(term()) :: term()
+  def inscpect(term) do
```
jonatanklosko, line 144: We already have `configure/1` for setting options, so not sure if we want options here as well?

josevalim: We may want options like `structs: false` when debugging certain scenarios.

lib/kino.ex
```diff
@@ -116,20 +116,35 @@ defmodule Kino do
...
+    output = Kino.Output.inspect(term)
+    Kino.Bridge.put_output(output)
+    term
```
josevalim, line 147: Can we add some basic tests? We can probably use `capture_io` to check that it is inspected and then itself is returned.

jonatanklosko: `capture_io` only captures to string, but we would need to mock the group leader to handle various messages and that actually tests `Kino.Bridge` instead. We may need some abstraction for testing the bridge since it started to grow, but maybe handle this separately?

josevalim: I think only capturing the string and not the registration would be fine as a starting point. Especially because we need to start testing which parts of Kino works with regular IO group leader, beyond the mock group leader too. So I would add the io group leader test now, do the mocking later.

lib/kino.ex
```diff
@@ -116,20 +116,39 @@ defmodule Kino do
...
+
+  @doc """
+  Inspects the given term as cell output.
+
+  This works essentially the same as `IO.inspect/2`, except it
+  always produces colored text and respects the configuration
+  set with `configure/1`.
+  """
+  @spec inscpect(term(), keyword()) :: term()
+  def inscpect(term, opts \\ []) do
```
josevalim, line 144:
```suggestion
  @spec inspect(term(), keyword()) :: term()
  def inspect(term, opts \\ []) do
```

lib/kino.ex
```diff
@@ -116,20 +116,39 @@ defmodule Kino do
...
-    nothing()
+    term
+  end
+
+  @doc """
+  Inspects the given term as cell output.
+
+  This works essentially the same as `IO.inspect/2`, except it
+  always produces colored text and respects the configuration
+  set with `configure/1`.
```
josevalim, line 141:
```suggestion
  set with `configure/1`.

  Opposite to `render/1`, it does not attempt to render the given
  term as a widget.
```

lib/kino.ex
jonatanklosko: My point is that capturing just the output doesn't work in this case, since we send a `{:livebook_put_output, _}` request to the group leader.

jonatanklosko: @josevalim I've just pushed a simple test for now, let me know what you think :)

josevalim: @jonatanklosko that's what I mean though, Kino.inspect should totally work even on a regular group leader. :)

josevalim: But yeah, we can work on making this work later. :)

josevalim (approved):

lib/kino.ex
jonatanklosko: That's something separate though, we are yet to add the terminal compatibility layer. It would basically convert all outputs to text or ignored, while currently it is `{:text, _}`.

PR #62, opened 2021-12-10
Add Kino.Control.form/2

jonatanklosko:
Adds support for form control. The form is composed of regular inputs, but inside form those inputs are not synchronized across users. Similarly, refreshing the page discards input values. As with other controls, there is no API for reading the value, instead we can subscribe to events, in this case submit or change. This enables each user to send inputs data a user-specific an event.

https://user-images.githubusercontent.com/17034772/145636268-e681bf03-cec6-4d89-bca8-221b19d86b6d.mp4

*For the Livebook part see https://github.com/livebook-dev/livebook/pull/790.*

### DIFF
--- a/lib/kino/control.ex
+++ b/lib/kino/control.ex
@@ -194,10 +194,10 @@ defmodule Kino.Control do
     end
 
     fields =
-      Keyword.map(fields, fn {_field, input} ->
+      Enum.map(fields, fn {field, input} ->
         # Make sure we use this input only in the form and nowhere else
         input = Kino.Input.duplicate(input)
-        input.attrs
+        {field, input.attrs}
       end)
 
     submit = Keyword.get(opts, :submit, nil)
--- a/lib/kino/control.ex
+++ b/lib/kino/control.ex
@@ -119,14 +119,14 @@ defmodule Kino.Control do
   Consequently, the form is another control for producing user-specific
   events.
 
-  Either `:submit` or `:report_change` must be specified.
+  Either `:submit` or `:report_changes` must be specified.
 
   ## Options
 
     * `:submit` - specifies the label to use for the submit button
       and enables submit events
 
-    * `:report_change` - whether to send new form value whenever any
+    * `:report_changes` - whether to send new form value whenever any
       of the input changes. Defaults to `false`
 
     * `:reset_on_submit` - a list of fields to revert to their default
@@ -144,12 +144,16 @@ defmodule Kino.Control do
 
   ## Examples
 
-  Create a form of out inputs:
+  Create a form out of inputs:
 
-      form = Kino.Control.form([
-        name: Kino.Input.text("Name"),
-        message: Kino.Input.textarea("Message")
-      ])
+      form =
+        Kino.Control.form(
+          [
+            name: Kino.Input.text("Name"),
+            message: Kino.Input.textarea("Message")
+          ],
+          submit: "Send"
+        )
 
   Subscribe to events:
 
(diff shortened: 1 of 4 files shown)


lib/kino/control.ex
```diff
@@ -111,6 +111,114 @@ defmodule Kino.Control do
...
+  events.
+
+  Either `:submit` or `:report_change` must be specified.
+
+  ## Options
+
+    * `:submit` - specifies the label to use for the submit button
+      and enables submit events
+
+    * `:report_change` - whether to send new form value whenever any
```
josevalim, line 129: Maybe `report_changes`? Also, what do you think about making the same as `reset_on_submit` and making it a list of all fields? Sometimes you may be interested only in certain changes? 🤔

lib/kino/control.ex
```diff
@@ -111,6 +111,114 @@ defmodule Kino.Control do
...
+    * `:data` - a map with field values, matching the field list
+
+  ## Examples
+
+  Create a form of out inputs:
+
+      form = Kino.Control.form([
+        name: Kino.Input.text("Name"),
+        message: Kino.Input.textarea("Message")
+      ])
```
josevalim, line 152: you forgot submit or report_changes ;)

lib/kino/control.ex
```diff
@@ -111,6 +111,114 @@ defmodule Kino.Control do
...
+  In addition to standard properties, all events include additional
+  properties.
+
+    * `:type` - either `:submit` or `:change`
+
+    * `:data` - a map with field values, matching the field list
+
+  ## Examples
+
+  Create a form of out inputs:
```
josevalim, line 147:
```suggestion
  Create a form out of inputs:
```

lib/kino/control.ex
jonatanklosko: Neat! In that case should we send all fields, listed fields or only the changed fields?

lib/kino/input.ex
```diff
@@ -39,6 +39,13 @@ defmodule Kino.Input do
     %__MODULE__{attrs: attrs}
   end
 
+  @doc false
+  def duplicate(input) do
+    input.attrs
+    |> Map.drop([:ref, :id, :destination])
+    |> new()
```
josevalim, line 46: I am worried about this being fragile in the long term because `new()` may not handle all of its attrs. What about instead of calling `new()`, make both `duplicate` and `new` call an internal function that sets those fields?

josevalim (commented):
Looks great, I have added some minor stuff!

lib/kino/input.ex
jonatanklosko: >  because new() may not handle all of its attrs

wdym? All other functions like `textarea` call new and it adds those three fields. Alternatively we could drop those fields directly in new and then do `duplicate(input), do: new(input.attrs)`?

lib/kino/control.ex
josevalim: Gah, this is tricky. I can all cases being useful. For example, for a chat app, we want to know if the message is being typed without necessarily sending it. So in this case, we want to track but not send it!

We could do `[field: true | :only_send | :only_track | false]`, where true stands for send and track, false is none, and we have two intermediate states. I don't think we need to implement this right now... but I want to make sure we avoid changes on the over the wire format. So maybe we keep only true/false in the public API but internally we send it as `field: true` for each field?

lib/kino/input.ex
josevalim: Ignore me. `new` is the function I was talking about. :)

lib/kino/control.ex
jonatanklosko: "Being typed" is tricky since we have debounce. How is `field: true` different from `field: :only_send`, I think both mean we should report the changed value?

jonatanklosko: Ah, I think I got it, send would attach to any change events, while track would actually trigger the events.

jonatanklosko: I made the over the wire format to include a map, since it probably makes more sense, currently only specifying true values.

josevalim (approved):

PR #64, opened 2021-12-23
Add support for defining custom JavaScript widgets

### LINKED ISSUE #44: [Question] - Steps to create new component
Hi!
I'm not sure where to direct my question so apologies if this isn't the correct forum or place. 

I'm trying to add a dynamic map component to Kino to interact with geographical data. It is loosely based on [Folium](http://python-visualization.github.io/folium/)

I'm trying to understand the flow and extension points in Kino and Livebook. 
My understanding so far:

### Kino
- add a new type and mapping in Kino.Output

- add a defimpl in Kino.Render (to_livebook)

- implement a server in Kino

### Livebook
- introduce a LiveComponent in Livebook for rendering the output

- create a hook for the client side JS part of it

- add a type in Livebook.Notebook.Cell.Elixir

Are there any other points to consider? 

Also I would like to thank all of your building such a fantastic tool! Keep up the good work!

jonatanklosko:
This introduces JavaScript powered widgets, which allow users to build arbitrary output visualizations and interactions.

It is an extremely exciting change, because it adds a powerful extension point to Livebook. To give more context, so far Livebook defined very specific output types that it supported, so for Vega-Lite we would have a dedicated output type and would include the necessary JavaScript in Livebook. With this new feature we can effectively remove all Vega-Lite specific bits from Livebook, because it is easily implemented within the new abstraction. Similarly, users can build integrations with arbitrary JavaScript packages without Livebook knowing anything about those.

Kino provides abstractions for defining those JavaScript widgets, both static (with static data) and live (with server process). Live widgets support bidirectional communication between the JavaScript part and the Elixir part.

To learn about the API and examples read `Kino.JS`, then `Kino.JS.Live`, then the linked notebook.

https://user-images.githubusercontent.com/17034772/147173550-5849dc02-4142-4f3d-a525-5d2ba089acd4.mp4

*Notebook available [here](https://gist.github.com/jonatanklosko/048171bb031370c02e4398f077da9fcb).*

Closes #44.

Also see https://github.com/livebook-dev/livebook/pull/818.


lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+  """
+  @callback init(arg :: term(), ctx :: Context.t()) :: {:ok, ctx :: Context.t()}
+
+  @doc """
+  Invoked whenever a new client connects to the server.
+
+  The returned data is passed to the JavaScript `init` function
+  of the connecting client.
+  """
+  @callback handle_connect(ctx :: Context.t()) :: {:ok, data :: term(), ctx :: Context.t()}
```
jonatanklosko, line 154: I'm thinking we should probably make `origin` available in `handle_connect` and `handle_event` to have a way of distinguishing clients. It could be an argument or a map with metadata (in case we want to add more info in the future). Thoughts?

josevalim:
This is amazing!!! The example notebook is fantastic!!!!!

lib/kino/js.ex
```diff
@@ -0,0 +1,362 @@
...
+  further interaction with the Elixir side. To introduce such
+  interaction, see `Kino.JS.Live` as a next step in our discussion.
+  '''
+
+  defstruct [:module, :data]
+
+  @type t :: %__MODULE__{module: module(), data: term()}
+
+  defmacro __using__(opts) do
+    quote location: :keep, bind_quoted: [opts: opts] do
```
josevalim, line 156: Ideally speaking you want to keep `quote` blocks as short as possible. In this case, I would do this:

```
quote do
  import Kino.JS, only: [asset: 2]
  @before_compile Kino.JS
  Kino.JS.__register__(assets_path)
end
```

`__register__` would do the job of calling `Module.put_attribute` to set external resources and default attributes.

lib/kino/js.ex
```diff
@@ -0,0 +1,362 @@
...
+          end
+
+          loaded_assets
+
+        {true, true} ->
+          IO.warn(
+            "ignoring files in #{inspect(assets_path)} because #{inspect(env.module)}" <>
+              " already defines inline assets with the assets/2 macro",
+            Macro.Env.stacktrace(env)
+          )
```
josevalim, line 245: Is there a reason for not supporting both given how it is implemented?

lib/kino/js.ex
```diff
@@ -0,0 +1,362 @@
...
+
+    :ok = :erl_tar.create(archive_path, archive_content, [:compressed])
+
+    {archive_path, hash}
+  end
+
+  defp assets_hash(assets) do
+    md5_hash =
+      assets
+      |> Enum.sort_by(fn {filename, _content} -> filename end)
```
josevalim, line 339:
```suggestion
      |> Enum.sort()
```

Given filenames can't be duplicate (and if they are, you want to order them somehow anyway).

lib/kino/js.ex
jonatanklosko: Beautiful, I didn't consider we can just pass the env :D

lib/kino/js.ex
jonatanklosko: Nothing in particular, it's mostly for consistency, as there's no reason to use both?

lib/kino/js.ex
```diff
@@ -0,0 +1,362 @@
...
+      |> Enum.flat_map(&Tuple.to_list/1)
+      |> :erlang.md5()
+
+    Base.encode32(md5_hash, case: :lower, padding: false)
+  end
+
+  defp dir_for_module(module) do
+    priv_dir = :code.priv_dir(:kino)
```
josevalim, line 347: Perhaps it is best to ask the user to pass the curernt otp app rather than put everything in Kino? Or is the plan exactly to put it all in Kino to make it easier to read later on?

josevalim: It seems you store an absolute path, so I would definitely do this:

    assets_path: {:current_app, "assets/kino"}

And, after further consideration, I would keep it in lib by default. `assets` directory is not pushed to `hex.pm` by default, so that will bring complication when it comes to publishing. So this:

    assets_path: {:current_app, "lib/mermaid-assets")

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+  As for the new bits, we added `use Kino.JS.Live` to define
+  a live widget server. We use `Kino.JS.Live.new/2` for creating
+  the widget instance and we implement a few `GenServer`-like
+  callbacks.
+
+  Once the widget server is started with `Kino.JS.Live.new/2`,
+  the `c:init/2` callback is called with the initial argument.
+  In this case we store the given `html` in server state.
+
+  Whenever the widget is rendered on the client, the `c:handle_connect/1`
```
josevalim, line 79:
```suggestion
  Whenever the widget is rendered on a new client, the `c:handle_connect/1`
```

lib/kino/js/live.ex
josevalim: Map with metadata sounds good to me.

josevalim: Another option is to have this information in the context, so you don't need to pass extra arguments on `handle_connect` and `handle_event`.

lib/kino/js.ex
jonatanklosko: To be clear, the Kino priv directory is for storing the generated archives, this way we ensure that after compilation all archives are in the Kino priv directory and would be correctly copied to a release, right? Also when defining widgets directly in Livebook, this way we have a known location when the archives would go.

As for the assets path, it currently requires the absolute path yeah, what would `:current_app` resolve to?

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+  See `c:GenServer.handle_cast/2` for more details.
+  """
+  @callback handle_cast(msg :: term(), ctx :: Context.t()) :: {:noreply, ctx :: Context.t()}
+
+  @doc """
+  Invoked to handle synchronous `call/3` messages.
+
+  See `c:GenServer.handle_call/3` for more details.
+  """
+  @callback handle_call(msg :: term(), {pid(), reference()}, ctx :: Context.t()) ::
```
josevalim, line 174: It is actually not guaranteed to be `{pid(), reference()}`. For example, in some cases it is `[alias | reference()]`. This should be an opaque term:

```suggestion
  @callback handle_call(msg :: term(), term(), ctx :: Context.t()) ::
```

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+  """
+  @callback handle_info(msg :: term(), ctx :: Context.t()) :: {:noreply, ctx :: Context.t()}
+
+  @doc """
+  Invoked when the server is about to exit.
+
+  See `c:GenServer.terminate/2` for more details.
+  """
+  @callback terminate(reason, ctx :: Context.t()) :: term()
+            when reason: :normal | :shutdown | {:shutdown, term()}
```
josevalim, line 190: If trapping exits the reason will be anything.

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+                      handle_call: 3,
+                      handle_cast: 2,
+                      handle_info: 2,
+                      terminate: 2
+
+  defmacro __using__(opts) do
+    quote location: :keep, bind_quoted: [opts: opts] do
+      @behaviour Kino.JS.Live
+
+      import Kino.JS.Live.Context, only: [assign: 2, update: 3, broadcast_event: 3]
```
josevalim, line 203: I would rather flip things around. Keep `assign/2`, `update/3` and user facing API in `Kino.JS.Live` and have the server in `Kino.JS.LiveServer`. This way we reduce the surface API for users.

lib/kino/js/live.ex
jonatanklosko: Interesting idea, my only concern would be that conceptually `ctx` is information about the server state, which is shared across all clients. But then it's already called context so it can be anything :see_no_evil:

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+    end
+  end
+
+  def __before_compile__(env) do
+    unless Module.defines?(env.module, {:__js_info__, 0}) do
+      message = """
+      make sure to include Kino.JS in #{inspect(env.module)} and define the necessary assets.
+
+          use Kino.JS
+          use Kino.JS.Live
```
josevalim, line 215: Is there any reason to tell them to `use Kino.JS.Live`? Might as well tell them to `use Kino.JS` but then start it with `Kino.JS.Live`. It is up to them to import the assign functions and friends.

lib/kino/js/live.ex
```diff
@@ -0,0 +1,369 @@
...
+      end
+
+    {:ok, %{module: module, client_pids: [], client_monitor_refs: [], ctx: ctx}}
+  end
+
+  @impl true
+  def handle_cast(msg, state) do
+    unless has_function?(state.module, :handle_cast, 2) do
+      raise "attempted to cast #{inspect(__MODULE__)}, but no handle_cast/2 was defined #{inspect(state.module)}"
+    end
```
josevalim, line 281: You don't need this, you can just call it and let it crash!

lib/kino/js/live.ex
jonatanklosko: Fair point, I was trying to stress out that `Kino.JS.Live` doesn't automatically inject `Kino.JS`, but juse `use Kino.JS` sounds good to me!

lib/kino/js/live/context.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  The event is dispatched to the registered JavaScript callback
+  on all connected clients.
+
+  ## Examples
+
+      broadcast_event(ctx, "new_point", %{x: 10, y: 10})
+  """
+  def broadcast_event(%__MODULE__{} = ctx, event, payload \\ nil) when is_binary(event) do
+    update_in(ctx, [Access.key!(:events)], &[{event, payload} | &1])
```
josevalim, line 56: Is there any reason to have this? Instead we could push the event directly the moment broadcast is called. That's how Phoenix Channels work. LiveView do not work like that because they always need to render.

lib/kino/js.ex
jonatanklosko: We still need to define `__mix_recompile__?`, so something like this:

```elixir
    quote location: :keep, bind_quoted: [opts: opts] do
      import Kino.JS, only: [asset: 2]

      @before_compile Kino.JS

      assets_path = opts[:assets_path]
      asset_paths = Kino.JS.__paths__(assets_path)
      Kino.JS.__register__(__ENV__, assets_path, asset_paths)

      # Force recompilation if new assets are added
      def __mix_recompile__? do
        current_paths = Kino.JS.__paths__(unquote(assets_path))
        :erlang.md5(current_paths) != unquote(:erlang.md5(asset_paths))
      end
    end
```

Any way to simplify this further?

josevalim: You can also define it in `__before_compile__` but your approach is also great!

lib/kino/js/live/context.ex
jonatanklosko: Given that `Context` is exposed to the user, it seems cleaner to store events there than private information like `client_pids`. Either approach works and I kinda like this more, any reason to not do that? (also that's something we can always change without affecting the API)

josevalim: The current API makes the cases where you need to do multiple sends harder. Imagine you want to send an initial payload, compute something, and send another payload. The only way to do so is by returning from the callback, send myself a message, and handling it in another callback.

jonatanklosko: I see, thanks!

lib/kino/js/live.ex
jonatanklosko: Sounds good, let me know if I interpreted it right c8345e2e8420a9e81ab698f0249e63569e12e706.

josevalim: Yes, exactly. Next I would remove `use Live` altogether.

jonatanklosko: The `@behaviour` + `import` combination is gonna be pretty much the default, shouldn't we make this scenario simpler?

josevalim: Ok then. Should it use Kino.JS automatically as well?

jonatanklosko: That's what I thought initially, but `Kino.JS` accepts options (`:asset_path`) and option forwarding feels too indirect.

jeregrine:
This is amazing! Great work!

lib/kino/js/live.ex
josevalim: Any way is fine by me. Just adding options. IMO it is fine for `ctx` to have the information about what triggered the callback. :D

lib/kino/js.ex
josevalim: Your call!

lib/kino/js.ex
josevalim: I see. There is indeed the case with no apps. `:current_app` is no longer relevant if we are always storing in Kino so ignore this issue. Although I would change `assets/mermaid` to `lib/...` something.

jonatanklosko: For Vega-Lite we do `use Kino.JS, assets_path: Path.expand("assets/vega_lite", __DIR__)`, so we target `lib/assets/vega_lite`. Or are you saying we should expand for the user somehow?

josevalim: Oh. I think for clarity you can do “lib/assets/vega_lite”, no need for expand. It will likely make the whole path clearer.

jonatanklosko: How do we get the root URL of the defining package? (and what about in Livebook?)

josevalim (approved):
LGTM! We can add the export feature in a following PR?

lib/kino/js/live.ex
jonatanklosko: Ah, one thing is that for callbacks other than `handle_connect` and `handle_event` it would be `nil`, but that's probably fine. In the end it's only useful in specific cases, so I like not needing to match-ignore it most of the time :)

lib/kino/js.ex
josevalim: The package is always compiled from its root.

jonatanklosko:
> We can add the export feature in a following PR?

Sounds good!

lib/kino/js.ex
jonatanklosko: Perfect! Just wasn't sure if that's an assumption we can make :D

lib/kino/js/live.ex
jonatanklosko: Added `ctx.origin` in 147be9334fe06ca065c940b8f301c870ee4ece15.

PR #65, opened 2021-12-27
Add support for exporting JS output

jonatanklosko:
This adds configuration options to enable exporting `Kino.JS` widgets into Live Markdown. Similar to what we already did for Vega-Lite, just generalized.

### DIFF
--- a/lib/kino/js.ex
+++ b/lib/kino/js.ex
@@ -382,6 +382,19 @@ defmodule Kino.JS do
     export =
       if info_string = opts[:export_info_string] do
         export_key = opts[:export_key]
+
+        if export_key do
+          unless is_map(data) do
+            raise ArgumentError,
+                  "expected data to be a map, because :export_key is specified, got: #{inspect(data)}"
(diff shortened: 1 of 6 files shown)


lib/kino/js.ex
```diff
@@ -352,9 +350,50 @@ defmodule Kino.JS do
...
   """
-  @spec new(module(), term()) :: t()
-  def new(module, data) do
-    %__MODULE__{module: module, data: data}
+  @spec new(module(), term(), keyword()) :: t()
+  def new(module, data, opts \\ []) do
+    export =
+      if info_string = opts[:export_info_string] do
+        export_key = opts[:export_key]
+        %{info_string: info_string, key: export_key}
```
josevalim, line 385: Should check if key exists on data?

josevalim (approved):

lib/kino/js.ex
jonatanklosko: Sounds good!

PR #68, opened 2022-01-03
Implement tables with Kino.JS.Live

jonatanklosko:
This changes the underlying implementation of `Kino.DataTable`, `Kino.ETS` and `Kino.Ecto` to JavaScript widgets, instead of the specific `:data_table` output we currently have. This removes another bit from the Kino-Livebook contract, which also means we can more easily change the tables behaviour in the future.

The public API doesn't change. The only difference is that the table state is now synchronized across uses, which we intended to do anyway.

https://user-images.githubusercontent.com/17034772/147990052-fb3c21f3-60f5-46d6-bbd0-b9a52e826540.mp4

I introduced a `Kino.Table` abstraction that builds on top of `Kino.JS.Live`, but I'd keep it private for now.

### DIFF
--- a/lib/assets/data_table/main.css
+++ b/lib/assets/data_table/main.css
@@ -1,0 +1,187 @@
+.app {
+  font-family: "Inter";
+
+  padding: 0px 12px 12px 12px;
+
+  --gray-50: #f8fafc;
+  --gray-100: #f0f5f9;
+  --gray-200: #e1e8f0;
+  --gray-300: #cad5e0;
+  --gray-400: #91a4b7;
+  --gray-500: #61758a;
+  --gray-700: #304254;
+  --gray-800: #1c2a3a;
+}
+
+.navigation {
+  margin-bottom: 16px;
+  display: flex;
+  align-items: center;
+}
+
+.navigation > :not(:first-child) {
+  margin-left: 12px;
+}
+
+.navigation__name {
+  font-size: 1rem;
+  font-weight: 600;
+  color: var(--gray-800);
+}
+
+.navigation__space {
+  flex-grow: 1;
+}
+
+.pagination {
+  display: flex;
+}
+
+.pagination > :not(:first-child) {
+  margin-left: 8px;
+}
+
+.pagination__button {
+  font-family: inherit;
+  background: none;
+  border: none;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  font-weight: 500;
+  font-size: 0.875rem;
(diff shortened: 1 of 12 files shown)


lib/assets/data_table/main.js
```diff
@@ -0,0 +1,149 @@
+import * as Vue from "https://cdn.jsdelivr.net/npm/vue@3.2.26/dist/vue.esm-browser.prod.js";
+
+export function init(ctx, data) {
+  ctx.importCSS("main.css");
+  ctx.importCSS("https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap");
+  ctx.importCSS("https://cdn.jsdelivr.net/npm/remixicon@2.5.0/fonts/remixicon.min.css");
+
+  const app = Vue.createApp({
```
jonatanklosko, line 8: As for the JavaScript part, I started with re-rendering the content with `innerHTML` (which is fairly ok in this case, as we don't render that much), however this resets all elements, in particular horizontal scroll position. We could use morphdom to optimise the DOM patching, but then I thought we may as well use something a bit more friendly to make the overall implementation more robust. I went with Vue, since it's easy to drop into a single file and the template doesn't require any fancy transpilation. I think it came out pretty neat and makes for a good example too.

lib/assets/data_table/main.css
```diff
@@ -0,0 +1,187 @@
...
+  padding: 0px 12px 12px 12px;
+
+  --gray-50: #f8fafc;
+  --gray-100: #f0f5f9;
+  --gray-200: #e1e8f0;
+  --gray-300: #cad5e0;
+  --gray-400: #91a4b7;
+  --gray-500: #61758a;
+  --gray-700: #304254;
+  --gray-800: #1c2a3a;
```
jonatanklosko, line 13: Since the page is sandboxed we need to replicate some Livebook-specific CSS, but it's not that much. In the future we could have a tiny Livebook CSS package for people to use.

josevalim (approved):
LGTM! Quick question: we can't merge Table and DataTable into a single module?

jonatanklosko:
> Quick question: we can't merge Table and DataTable into a single module?

Technically we could, because DataTable only implements the behaviour, but the same goes for ETS and Ecto. From a logical point of view it would be very weird, no?

josevalim:
Oh, ignore me. I thought that DataTable was internal but it is not. It is what we used for rendering any data. Ship it!

PR #70, opened 2022-01-05
Add reference to JavaScript outputs and messages

jonatanklosko:
Adds a reference to JS widgets and the widget messages. This way we can do multiplexing with a single subscription process in https://github.com/livebook-dev/livebook/pull/843.


josevalim (approved):

jonatanklosko:
@josevalim I've updated this one too, let me know if you have any further comments :D

josevalim (approved):

PR #71, opened 2022-01-07
Introduce broadcast communication

jonatanklosko:
Adds `Kino.Bridge.broadcast(topic, message)` that goes through the group leader and is later broadcasted to interested processes on Livebook side. This way we can send events from JS server to Livebook as a single message and use the pubsub there, as opposed to sending the event to individual processes in Livebook.

Also adds `Kino.Bridge.send(pid, message)`, so that all communication goes through the bridge API.

### DIFF
--- a/lib/kino/js/live_server.ex
+++ b/lib/kino/js/live_server.ex
@@ -17,7 +17,7 @@ defmodule Kino.JS.LiveServer do
 
   def broadcast_event(ctx, event, payload) do
     ref = ctx.__private__.ref
-    Kino.Bridge.broadcast("js_live:#{ref}", {:event, event, payload, %{ref: ref}})
+    Kino.Bridge.broadcast("js_live", ref, {:event, event, payload, %{ref: ref}})
     :ok
   end
 
--- a/lib/kino/js/live_server.ex
+++ b/lib/kino/js/live_server.ex
@@ -16,10 +16,8 @@ defmodule Kino.JS.LiveServer do
   defdelegate call(pid, term, timeout), to: GenServer
 
   def broadcast_event(ctx, event, payload) do
-    for pid <- ctx.__private__.client_pids do
-      send(pid, {:event, event, payload, %{ref: ctx.__private__.ref}})
-    end
-
+    ref = ctx.__private__.ref
+    Kino.Bridge.broadcast("js_live:#{ref}", {:event, event, payload, %{ref: ref}})
     :ok
   end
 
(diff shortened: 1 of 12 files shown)


lib/kino/js/live_server.ex
```diff
@@ -16,10 +16,8 @@ defmodule Kino.JS.LiveServer do
   defdelegate call(pid, term, timeout), to: GenServer
 
   def broadcast_event(ctx, event, payload) do
-    for pid <- ctx.__private__.client_pids do
-      send(pid, {:event, event, payload, %{ref: ctx.__private__.ref}})
-    end
-
+    ref = ctx.__private__.ref
+    Kino.Bridge.broadcast("js_live:#{ref}", {:event, event, payload, %{ref: ref}})
```
jonatanklosko, line 20: I went with a more specific topic and allowed subscribing to `js_live:*` on Livebook side. In the channel we want all of those for the given session, but I can imagine subscribing to specific output only.

jonatanklosko: My only concern is that this obfuscates what topic we subscribe to on Livebook side, but that's internal so it's probably fine. Does it still makes sense if we subscribe to `js_live:<ref>` in the channel though?

josevalim: I would separate just to give us more control. Think about it like this: the fact it uses a pubsub on the livebook server should be an implementation detail from Kino perspective. So the arguments should not necessarily be mirror after pubsub.

jonatanklosko: Sounds good!

jonatanklosko:
Added a Livebook-like group leader for tests, so that we can tests parts relying on `Kino.Bridge`. Here specifically we want to assert that the events are broadcasted (while previously we would just subscribe the test process to the widget directly). Also added a number of functions shared across the `Kino.JS.Live` widgets, we can make it public once things settle (we will also need a clean way for people to use the new group leader for those tests).

test/support/livebook_case.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Kino.LivebookCase do
+  use ExUnit.CaseTemplate
+
+  setup do
+    original_gl = Process.group_leader()
+
+    gl = start_supervised!({KinoTest.GroupLeader, self()})
+    Process.group_leader(self(), gl)
+
+    on_exit(fn ->
```
josevalim, line 10: I don’t think you need to revert because the process will exit anyway?

josevalim (approved):

test/support/livebook_case.ex
jonatanklosko: Right, I did it just in case, but all regular requests are forwarded to the regular group leader, so it's definitely fine to remove this either way :)

PR #75, opened 2022-01-12
Restructure frame output

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/857.


test/support/livebook_case.ex
```diff
@@ -1,6 +1,12 @@
 defmodule Kino.LivebookCase do
   use ExUnit.CaseTemplate
 
+  using do
+    quote do
+      import KinoTest.Livebook
```
josevalim, line 6: FWIW, you can have the helper functions in this module and then import this module itself.

josevalim (approved):

PR #76, opened 2022-01-17
Add stream API for control events

jonatanklosko:
https://user-images.githubusercontent.com/17034772/149792089-d69a02b4-c413-46f5-9604-3f7a28075a9f.mp4

### DIFF
--- a/lib/kino/subscription_manager.ex
+++ b/lib/kino/subscription_manager.ex
@@ -60,6 +60,7 @@ defmodule Kino.SubscriptionManager do
       fn ref ->
         receive do
           {^ref, event} -> {[event], ref}
+          {:__topic_cleared__, ^topic, ^ref} -> {:halt, ref}
         end
       end,
       fn _ref -> unsubscribe(topic, self()) end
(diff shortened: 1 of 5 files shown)


lib/kino/subscription_manager.ex
```diff
@@ -92,7 +113,8 @@ defmodule Kino.SubscriptionManager do
     {subscribers, state} = pop_in(state.topic_with_subscribers[topic])
 
     state =
-      Enum.reduce(subscribers || [], state, fn {pid, _tag}, state ->
+      Enum.reduce(subscribers || [], state, fn {pid, tag}, state ->
+        send(pid, {:__topic_cleared__, topic, tag})
```
josevalim, line 117: Isn't this going to send it for all subscribers? Even the ones that are not streams? 🤔

jonatanklosko: Yeah, alternatively we can subscribe with a special tag like `{:__stream__, ref}` and send message only for those tags, wdyt?

josevalim: Either way is fine. We can send it to everyone, but if we do, then we need to document it. Weren't we planning on having join and leave events anyway?

jonatanklosko: The join/leave events concern the user, not the widget disappearing. I think it makes more sense to message the stream only, because otherwise we may spam the subscribing process and there's likely nothing it can do, just waits for cell reevaluation.

josevalim: But for keyboard, we want to know when the keyboard is deactivated and the user leaves the page, right? In any case, keeping this only for streams for now is a great starting point.

jonatanklosko: For keyboard the user decides if they want the `:status` events, if so they receive event on activation/deactivation. It's different from keyboard being removed on cell reevaluation though.

jonatanklosko: Done!

josevalim (approved):

PR #77, opened 2022-01-17
Add type validations for default input value

jonatanklosko:
image

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -56,7 +56,14 @@ defmodule Kino.Input do
   @spec text(String.t(), keyword()) :: t()
   def text(label, opts \\ []) when is_binary(label) and is_list(opts) do
     default = Keyword.get(opts, :default, "")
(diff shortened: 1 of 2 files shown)


lib/kino/input.ex
```diff
@@ -56,9 +56,16 @@ defmodule Kino.Input do
   @spec text(String.t(), keyword()) :: t()
   def text(label, opts \\ []) when is_binary(label) and is_list(opts) do
     default = Keyword.get(opts, :default, "")
+    assert_default_value!(default, "be a string", &is_binary/1)
```
jonatanklosko, line 59: @josevalim let me know if you'd convert instead, but I'm thinking being explicit makes sense.

jonatanklosko: I see, let's do that!

jonatanklosko:
Updated!

josevalim (approved):

PR #78, opened 2022-01-18
Add combined stream for multiple controls

jonatanklosko:
https://user-images.githubusercontent.com/17034772/149975139-cc5dbdba-33eb-4dc6-874f-8f56962d7631.mp4

### DIFF
--- a/lib/kino/control.ex
+++ b/lib/kino/control.ex
@@ -83,7 +83,7 @@ defmodule Kino.Control do
 
   Create the widget:
 
-      keyboard = Kino.Control.keyboard([:keyup, :keydown])
+      keyboard = Kino.Control.keyboard([:keyup, :keydown, :status])
 
   Subscribe to events:
 
@@ -92,6 +92,7 @@ defmodule Kino.Control do
   As the user types events are streamed:
 
       IEx.Helpers.flush()
+      #=> {:keyboard, %{enabled: true, origin: #PID<10895.9854.0>, type: :status}
       #=> {:keyboard, %{key: "o", origin: #PID<10895.9854.0>, type: :keydown}}
       #=> {:keyboard, %{key: "k", origin: #PID<10895.9854.0>, type: :keydown}}
       #=> {:keyboard, %{key: "o", origin: #PID<10895.9854.0>, type: :keyup}}
@@ -257,6 +258,10 @@ defmodule Kino.Control do
   Returns a new interval event source.
 
   This can be used as event source for `stream/1` and `tagged_stream/1`.
+  The events are emitted periodically with an increasing value, starting
+  from 0 and have the form:
+
+      %{type: :interval, iteration: 0}
   """
   @spec interval(non_neg_integer()) :: interval()
   def interval(milliseconds) when is_number(milliseconds) and milliseconds > 0 do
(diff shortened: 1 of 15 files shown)


lib/kino/control.ex
```diff
@@ -34,6 +34,8 @@ defmodule Kino.Control do
 
   @type t :: %__MODULE__{attrs: Kino.Output.control_attrs()}
 
+  @type interval :: {:interval, milliseconds :: non_neg_integer()}
```
josevalim, line 37:
```suggestion
  @opaque interval :: {:interval, milliseconds :: non_neg_integer()}
```

lib/kino/control.ex
```diff
@@ -230,33 +232,169 @@ defmodule Kino.Control do
...
+    {:interval, milliseconds}
   end
 
   @doc """
   Returns a `Stream` of control events.
 
   This is an alternative API to `subscribe/2`, such that event
   messages are consume via stream instead of process messages.
+
+  Accepts a single source or a list of sources, where each source
```
josevalim, line 272:
```suggestion
  It accepts a single source or a list of sources, where each source
```

lib/kino/control.ex
```diff
@@ -230,33 +232,169 @@ defmodule Kino.Control do
...
+  is either of:
+
+    * `Kino.Control` - emitting value on relevant interaction
+
+    * `Kino.Input` - emitting value on value change
+
+    * `Kino.Control.interval(n)` - periodically emitting an
+      increasing value, starting from 0
```
josevalim, line 280:
```suggestion
```

lib/kino/control.ex
```diff
@@ -230,33 +232,169 @@ defmodule Kino.Control do
...
+
+  @doc """
+  Same as `stream/1`, but attaches custom tag to every stream item.
+
+  ## Example
+
+      button = Kino.Control.button("Hello")
+      input = Kino.Input.checkbox("Check")
+
+      for event <- Kino.Control.stream([hello: button, check: input]) do
```
josevalim, line 328:
```suggestion
      for event <- Kino.Control.tagged_stream([hello: button, check: input]) do
```

lib/kino/control.ex
```diff
@@ -230,33 +232,169 @@ defmodule Kino.Control do
...
+            {[mapper.(tag, event)], {ref, topics}}
+
+          {{^ref, _tag}, :topic_cleared, topic} ->
+            case topics -- [topic] do
+              [] -> {:halt, {ref, []}}
+              topics -> {[], {ref, topics}}
+            end
+
+          {{^ref, tag}, :__interval__, ms, n} ->
+            Process.send_after(self(), {{ref, tag}, :__interval__, ms, n + 1}, ms)
```
josevalim, line 389: I think the events emitted should be `%{type: :interval, iteration: n}` or something like that. Please make sure to document the type on `Kino.Control.interval`. :)

lib/kino/control.ex
jonatanklosko: I recently thought we should do that for all widget types, since the internal format shouldn't be relevant to the user, wdyt?

josevalim: Sounds good to me!

lib/kino/control.ex
jonatanklosko: I was keeping it in line with `Stream.interval/1`, but sounds better to align the events instead yeah!

lib/kino/control.ex
```diff
@@ -230,33 +233,171 @@ defmodule Kino.Control do
...
+  end
+
+  @doc """
+  Returns a new interval event source.
+
+  This can be used as event source for `stream/1` and `tagged_stream/1`.
+  The events are emitted periodically with an increasing value, starting
+  from 0 and have the form:
+
+      %{type: :interval, iteration: 0}
```
josevalim, line 264:
```suggestion
      %{type: :interval, iteration: integer()}
```

josevalim (approved):
one last comment and ship it!

lib/kino/control.ex
jonatanklosko, line 264:
```suggestion
      %{type: :interval, iteration: non_neg_integer()}
```

Issue #74, opened 2022-01-12
More tables more strict


josevalim:
Today tables are very flexible in that they can render mixed data types, but that's going to conflict with some of the features we want to add in the future. So I think we should do a more explicit modelling, where we choose the data type based on the first entry of the list, as well as the fields, and assume everything is there.

For ETS, we should have a single column, which is the row, and print them as tuples.

PR #87, opened 2022-02-07
Improves table accessibility

cristineguadelupe:
Increases the contrast of the table pagination buttons
Adds missing aria-labels 
Adds the `sr-only` class
Creates a live-region for sorting and pagination information

Video with the original voice over audio

https://user-images.githubusercontent.com/5660941/152708177-bf0935d7-65fd-4c9e-a680-a5b7897d87c2.mov

### DIFF
--- a/lib/assets/data_table/main.js
+++ b/lib/assets/data_table/main.js
@@ -26,6 +26,7 @@ export function init(ctx, data) {
           <div
             v-if="data.features.includes('pagination') && data.content.rows.length > 0"
             class="pagination"
+            aria-controls="table-info"
           >
             <button
               class="pagination__button"
@@ -68,7 +69,7 @@ export function init(ctx, data) {
                   class="table__cell"
                   :class="{ 'table__cell--clickable': data.features.includes('sorting') }"
                   @click="data.features.includes('sorting') && orderBy(column.key)"
-                  aria-label="click to order by"
+                  aria-controls="table-info"
                 >
                   <div class="table__cell-content">
                     <span>{{ column.label }}</span>
@@ -88,6 +89,10 @@ export function init(ctx, data) {
             </tbody>
           </table>
         </div>
+      </div>
+      <div class="sr-only" role="region" id="table-info" aria-live="assertive">
+        <span>{{ data.content.order_by ? orderedBy() : 'unordered'  }}</span>
+        <spa>you are on table page {{ data.content.page }}</spa>
       </div>
     `,
 
--- a/lib/assets/data_table/main.js
+++ b/lib/assets/data_table/main.js
@@ -60,7 +60,7 @@ export function init(ctx, data) {
           class="table-container tiny-scrollbar"
           :class="{ 'table-container--loading': loading }"
         >
-          <table class="table">
+          <table class="table" :aria-label="data.content.order_by ? orderedBy() : '' ">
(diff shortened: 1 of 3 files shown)


lib/assets/data_table/main.js
```diff
@@ -116,7 +123,14 @@ export function init(ctx, data) {
       orderBy(newKey) {
         this.loading = true;
         const [key, order] = reorder(this.data.content.order_by, this.data.content.order, newKey)
-        ctx.pushEvent("order_by", { key, order });
+        const label = key ? this.data.content.columns[key].label : null
```
jonatanklosko, line 126: Please let me know what I miss, but I think `content.columns` is an array, and `key` is a string, how does this work? Shouldn't we instead find a column with `column.key === key`?

lib/assets/data_table/main.js
```diff
@@ -116,7 +123,14 @@ export function init(ctx, data) {
...
+        ctx.pushEvent("order_by", { key, order, label});
```
jonatanklosko, line 127: It looks like we are sending the label to the server, just to get it back. Any reason we cannot compute the label based on `columns` and `order_by`?

lib/assets/data_table/main.js
```diff
@@ -116,7 +123,14 @@ export function init(ctx, data) {
...
+      },
+
+      orderedBy() {
```
jonatanklosko, line 130: Let's make it clearer that it returns a human-readable string :)

```suggestion
      orderDescription() {
```

lib/assets/data_table/main.js
cristineguadelupe: We can and we should! The reason is my habit of doing liveview things server-side only. Done :)

lib/assets/data_table/main.js
```diff
@@ -118,6 +125,13 @@ export function init(ctx, data) {
         const [key, order] = reorder(this.data.content.order_by, this.data.content.order, newKey)
         ctx.pushEvent("order_by", { key, order });
       },
+
+      orderDescription() {
+        const key = this.data.content.order_by
+        const label = this.data.content.columns[key].label
```
cristineguadelupe, line 131: I can change it if you prefer. But I'm basically doing both at once. Fetching the column from the array and accessing the corresponding label.
Now that I removed the conditional, this breaks if the key is not in the array, but this code is only called when we know the key is there.

cristineguadelupe: Or maybe I didn't pay enough attention and not always the key of the column will be its position… 🤔
My turn to be confused!

jonatanklosko: On the elixir side the keys can be of arbitrary data type, so we map them to strings when sending to the client. It happens that we generate those strings sequentially, so they make for valid indices in most cases. However, in cases like ETS the columns may change as we go through pages, and then we generate more keys and they are no longer valid indices. Either way, we shouldn't rely on what those keys are, so instead let's do `columns.find((column) => column.key === key)` :)

josevalim: Note there is an option issue #74 to make the tables less strict. In particular, should be printed as tables of a single column where the column is the value itself. This should simplify a lot of things internally. :) But for now we need to handle it, yeah. :)

PR #90, opened 2022-02-08
Renders ETS as tuples

cristineguadelupe:
This PR changes the way we show ETS tables. Now we always have only one column named `rows` and all rows are just tuples.

Screen Shot 2022-02-08 at 00 55 13

### DIFF
--- a/lib/kino/ets.ex
+++ b/lib/kino/ets.ex
@@ -12,8 +12,6 @@ defmodule Kino.ETS do
 
   @behaviour Kino.Table
 
-  alias Kino.Utils
-
   @type t :: Kino.JS.Live.t()
 
   @doc """
@@ -51,9 +49,9 @@ defmodule Kino.ETS do
   @impl true
   def get_data(rows_spec, state) do
     records = get_records(state.tid, rows_spec)
-    rows = Enum.map(records, &record_to_row/1)
+    rows = Enum.map(records, fn record -> %{fields: %{0 => inspect(record)}} end)
     total_rows = :ets.info(state.tid, :size)
-    columns = records |> Utils.Table.keys_for_records() |> Utils.Table.keys_to_columns()
+    columns = if records == [], do: [], else: [%{key: 0, label: "rows"}]
     {:ok, %{columns: columns, rows: rows, total_rows: total_rows}, state}
   end
 
@@ -68,15 +66,5 @@ defmodule Kino.ETS do
     records = :qlc.next_answers(cursor, rows_spec.limit)
     :qlc.delete_cursor(cursor)
     records
-  end
-
-  defp record_to_row(record) do
-    fields =
-      record
-      |> Tuple.to_list()
-      |> Enum.with_index()
-      |> Map.new(fn {val, idx} -> {idx, inspect(val)} end)
-
-    %{fields: fields}
   end
 end
--- a/test/kino/ets_test.exs
+++ b/test/kino/ets_test.exs
@@ -62,13 +62,12 @@ defmodule Kino.ETSTest do
     assert %{
              content: %{
                columns: [
-                 %{key: "0", label: "0"},
-                 %{key: "1", label: "1"}
+                 %{key: "0", label: "rows"}
                ],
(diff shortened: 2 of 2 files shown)


lib/kino/ets.ex
```diff
@@ -51,9 +49,9 @@ defmodule Kino.ETS do
   @impl true
   def get_data(rows_spec, state) do
     records = get_records(state.tid, rows_spec)
-    rows = Enum.map(records, &record_to_row/1)
+    rows = Enum.map(records, fn record -> %{fields: %{0 => inspect(record)}} end)
     total_rows = :ets.info(state.tid, :size)
-    columns = records |> Utils.Table.keys_for_records() |> Utils.Table.keys_to_columns()
+    columns = if records == [], do: [], else: [%{key: 0, label: "rows"}]
```
jonatanklosko, line 54: Since we know the table structure, we can always show the expected columns :) I think the column should be singular, like item/value/entry/record, any preference?

```suggestion
    columns = [%{key: 0, label: "item"}]
```

jonatanklosko: (we already use `:item` for Ecto results that are lists of values, we should make sure the wording is the same)

josevalim: I would prefer row in this case because it is the whole ETS row (not sure if it is the case for Ecto). :)

wojtekmach:
Given rows are always Elixir terms I wonder if it would make sense to always print them in `<pre>`? Maybe there should be an option to DataTable?

lib/kino/ets.ex
jonatanklosko: I see, for Ecto it's only the case when doing like `select: user.name`, so it's different. Let's go with "row" then!

cristineguadelupe: I'm not a huge fan of showing the header without data, but no strong opinion. It's up to you!
`:item` sounds good and keeps consistency. Like it!

josevalim: My suggestion:

```suggestion
    columns = [%{key: 0, label: "row", type: "tuple"}]
```

Am I using type correctly @jonatanklosko ?

jonatanklosko: @josevalim yup!

jonatanklosko: @cristineguadelupe if we want to hide columns on no data, we should do it in the template directly, so this applies to all cases. @josevalim any preference?

cristineguadelupe: What about just show "No data" like we are doing for other tables?

jonatanklosko:
@wojtekmach good call! Actually, I'm wondering if we should just always use monospaced font for the table, since we generally show inspected data. This way the styling would be consistent.

lib/kino/ets.ex
jonatanklosko: @cristineguadelupe exactly, my question is if we want to do that. If so, then we just change the condition on the front end, so we don't need to remember about it for every table.

josevalim:
@jonatanklosko I think monospaced functions will be bad for database data. We should definitely allow multiple representations there. It should be an option.

josevalim:
Monospaced won't be good for Explorer either. :)

lib/kino/ets.ex
cristineguadelupe: Gotcha! I think we should. Looks better, clearer, and consistent.

jonatanklosko:
@josevalim to be honest I think it looks better. Deepnote uses monospaced font for the tables too :)

image

image

lib/kino/ets.ex
josevalim: Sorry for ruining the party but I don't think we should. If you make a SQL query, for example, even if it returns no results, I think knowing the column names and types is important information. Same for when showing data frames.

cristineguadelupe: Ok, I made the other changes and kept the header

josevalim (approved):

jonatanklosko (approved):

PR #94, opened 2022-02-09
Simplify column computation and add types to Ecto results

jonatanklosko:
image

### DIFF
--- a/lib/kino/utils/table.ex
+++ b/lib/kino/utils/table.ex
@@ -62,9 +62,12 @@ defmodule Kino.Utils.Table do
   def columns_for_schema(schema) do
     for field <- schema.__schema__(:fields) do
       type = schema.__schema__(:type, field)
-      %{key: field, label: inspect(field), type: inspect(type)}
+      %{key: field, label: inspect(field), type: ecto_type_to_string(type)}
     end
   end
+
+  defp ecto_type_to_string({:parameterized, module, _info}), do: inspect(module)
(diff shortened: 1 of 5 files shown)


lib/kino/utils/table.ex
```diff
@@ -1,77 +1,75 @@
...
   end
 
+  @doc """
+  Computes table column specifications for the given Ecto schema.
+  """
+  @spec columns_for_schema(module()) :: list(Kino.Table.column())
+  def columns_for_schema(schema) do
+    for field <- schema.__schema__(:fields) do
+      type = schema.__schema__(:type, field)
+      %{key: field, label: inspect(field), type: to_string(type)}
```
josevalim, line 65: Composite and parameterized types can be tuples. :)

lib/kino/utils/table.ex
```diff
@@ -1,77 +1,75 @@
...
-        else
-          all_keys
-        end
+      nil -> []
+      first_record -> columns_for_record(first_record)
     end
   end
 
-  defp keys_for_record(record) when is_tuple(record) do
+  defp columns_for_record(record) when is_tuple(record) do
```
josevalim, line 24: This clause can be removed?

josevalim (commented):
Just one tiny comment about Ecto and ship it!

lib/kino/utils/table.ex
jonatanklosko: Good point, so in that case `inspect` should look fine, right?

image

PR #98, opened 2022-02-25
Add support for smart cell definitions

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/1029.

Currently this builds directly on top of `Kino.JS.Live`. A new cell is defined by implementing `Kino.JS.Live` and also `Kino.SmartCell`, which describes the state-code mapping. These APIs are likely to change as we extend smart cell capabilities.


lib/kino/smart_cell/server.ex
```diff
@@ -0,0 +1,118 @@
+defmodule Kino.SmartCell.Server do
```
jonatanklosko, line 1: This has a number of similarities to `Kino.JS.Live.Server` and we can easily extract some pieces, but for now I kept it separate to make further changes easier.

lib/kino/smart_cell/db_connection.ex
```diff
@@ -0,0 +1,298 @@
+defmodule Kino.SmartCell.DBConnection do
+  @moduledoc false
+
+  # A smart cell used to establish connection to a database.
+
+  use Kino.JS
+  use Kino.JS.Live
+  use Kino.SmartCell, name: "Database connection"
```
jonatanklosko, line 8: Should we make it `Kino.JS.SmartCell` for consistency? (though from Livebook perspective it may feel a bit arbitrary)

lib/kino/smart_cell.ex
```diff
@@ -0,0 +1,235 @@
...
+    name = Keyword.fetch!(opts, :name)
+
+    quote do
+      def child_spec(%{ref: ref, attrs: attrs, target_pid: target_pid}) do
+        spec = %{
+          id: __MODULE__,
+          start: {Kino.SmartCell.Server, :start_link, [__MODULE__, ref, attrs, target_pid]}
+        }
+
+        Supervisor.child_spec(spec, restart: :temporary)
```
josevalim, line 185:
```suggestion
          start: {Kino.SmartCell.Server, :start_link, [__MODULE__, ref, attrs, target_pid]},
          restart: :temporary
        }
```

lib/kino/smart_cell.ex
```diff
@@ -0,0 +1,235 @@
...
+
+  ## Example
+
+  As a minimal example, that's how we can define a cell that allows
+  editing the underlying code directly through a textarea.
+
+      defmodule Kino.SmartCell.Plain do
+        use Kino.JS
+        use Kino.JS.Live
+        use Kino.SmartCell, name: "Plain code editor"
```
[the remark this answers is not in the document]
jonatanklosko: We will have more attributes, some optional. For example we could have `eager: true` to indicate the code should be evaluated whenever it changes (for the graph generator).

We could have a single callback that returns the options, though with this approach we emit compile time errors, if attributes are missing.

Let me know what you prefer in this context :)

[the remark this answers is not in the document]
josevalim: Sounds good to me to keep it as is!

lib/kino/smart_cell/db_connection.ex
josevalim: `Kino.SmartCell` is fine IMO. :)

PR #99, opened 2022-03-04
Add callbacks for scanning evaluation context

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/1041.

### DIFF
--- a/lib/kino/smart_cell.ex
+++ b/lib/kino/smart_cell.ex
@@ -166,27 +166,21 @@ defmodule Kino.SmartCell do
   This callback receives the binding and environment available to the
   smart cell code.
 
-  Note that this callback runs asynchronously and the result is then
-  handled with `c:scan_binding_continue/2` where the server state can
-  be updated accordingly.
+  Note that this callback runs asynchronously and it receives the PID
+  of the smart cell server, so the result needs to be sent explicitly
+  and handled using `c:Kino.JS.Live.handle_info/2`.
 
-  **Important:** the return value is copied between processes, so
-  avoid returning large data structures. In particular, when looking
-  at variables, instead of returning their values, extract and return
+  **Important:** remember that data sent between processes is copied,
+  so avoid sending large data structures. In particular, when looking
+  at variables, instead of sending their values, extract and send
   only the relevant metadata.
 
   **Important:** avoid any heavy work in this callback, as it runs in
   the same process that evaluates code, so we don't want to block it.
   """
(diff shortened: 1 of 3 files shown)


lib/kino/smart_cell.ex
```diff
@@ -160,6 +160,31 @@ defmodule Kino.SmartCell do
...
+  at variables, instead of returning their values, extract and return
+  only the relevant metadata.
+  """
+  @callback scan_binding(Code.binding(), Macro.Env.t()) :: data :: term()
+
+  @doc """
+  Invoked with the result of `c:scan_binding/2` to update the server
+  state.
+  """
+  @callback scan_binding_continue(data :: term(), ctx :: Context.t()) :: ctx :: Context.t()
```
jonatanklosko, line 184: @josevalim we could similarly add `scan_evaluation_result` and `scan_evaluation_result_continue`. Implementation-wise I think it's pretty much the same as `handle_exception`. It's more verbose, but also gives more flexibility. Thoughts?

jonatanklosko: To only handle errors we could then do:

```elixir
@impl true
def scan_evaluation_result({:ok, _}), do: :none
def scan_evaluation_result({:error, kind, reason, stacktrace}), do: {kind, reason, stacktrace}

@impl true
def scan_evaluation_result_continue(exception) do
  # handle_exception
end
```

lib/kino/smart_cell.ex
```diff
@@ -160,6 +160,34 @@ defmodule Kino.SmartCell do
...
+  **Important:** avoid any heavy work in this callback, as it runs in
+  the same process that evaluates code, so we don't want to block it.
+  """
+  @callback scan_binding(Code.binding(), Macro.Env.t()) :: data :: term()
+
+  @doc """
+  Invoked with the result of `c:scan_binding/2` to update the server
+  state.
+  """
+  @callback scan_binding_continue(data :: term(), ctx :: Context.t()) :: ctx :: Context.t()
```
josevalim, line 187: For completeness, there is another API you can use here:

    scan_binding(pid, code, env)

Where `pid` is the `pid` of the smart cell. This way we have fewer callbacks and give a bit more flexibility to the user. I think I prefer this approach.

If you prefer two callbacks, then i would perhaps change the names to: `prepare_scan_binding` and `handle_scan_binding`. `handle_` is a known name for server callbacks after all and that would make it more apparent.

josevalim (approved):
LGTM! Just one tiny comment!

josevalim (approved):

PR #105, opened 2022-03-16
Add SQL query smart cell

jonatanklosko:
https://user-images.githubusercontent.com/17034772/158706294-9d804ef3-e0b3-4da5-be0a-5babc298773f.mp4


lib/kino/smart_cell/sql.ex
```diff
@@ -0,0 +1,200 @@
...
+      "result_variable" => ctx.assigns.result_variable
+    }
+  end
+
+  @impl true
+  def to_source(attrs) do
+    attrs
+    |> to_quoted()
+    |> Macro.to_string()
+    # TODO: remove reformatting on Elixir v1.14 (before v1.13.1
```
josevalim, line 117: Macro.to_string is using the same foundation as `Code.format_string` in Elixir v1.13, so you can use the direct APIs instead: https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/macro.ex#L971-L974

jonatanklosko:
@josevalim the default query. I'm not sure about capitalization, but for quick exploration I don't think people would capitalize the keywords, so we shouldn't either, wdyt?

https://user-images.githubusercontent.com/17034772/158807810-76c09b6a-8082-4079-ace1-993d8b95451a.mp4

josevalim:
@jonatanklosko lgtm!

josevalim (approved):

PR #106, opened 2022-03-17
Add Chart Builder

cristineguadelupe:
Initial implementation of Chart Builder

In this initial stage, we have a "UI to chart" where the goal is to allow the creation of the most used charts from a simple interface.

https://user-images.githubusercontent.com/5660941/158725512-88fbce34-ec3b-459a-bb71-ce2f9491a3c3.mp4

CleanShot 2022-03-16 at 23 40 33@2x

CleanShot 2022-03-16 at 23 40 46@2x

CleanShot 2022-03-16 at 23 13 06@2x


lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  defp to_quoted(attrs) do
+    data = attrs["data"]
+    module = if attrs["vl_alias"], do: attrs["vl_alias"], else: VegaLite
+    module = Module.split(module) |> hd() |> String.to_atom()
+
+    attrs =
+      attrs
+      |> Enum.map(fn {k, v} -> {String.to_atom(k), v} end)
+      |> Enum.into(%{})
```
polvalente, line 110:
```suggestion
      Map.new(attrs, fn {k, v} -> {String.to_atom(k), v} end)
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+
+    [{_, root} | nodes] = [
+      new: %{name: :new, module: module, args: [[width: attrs.width, height: attrs.height]]},
```
polvalente, line 113:
```suggestion
      %{field: :new, function_name: :new, module: module, args: [[width: attrs.width, height: attrs.height]]},
```

I was thinking about this data structure, and now it seems to me that abandoning the outer keyword in favor of a list of maps could be better.

Some matches below would also need to change, but it makes for a cleaner code overall. `[root | nodes] = ...` for instance

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+      mark: %{name: :mark, module: module, args: [attrs.chart]},
+      x: %{name: :encode_field, module: module, args: [attrs.x_axis, [type: attrs.x_axis_type]]},
+      y: %{name: :encode_field, module: module, args: [attrs.y_axis, [type: attrs.y_axis_type]]},
+      color: %{name: :encode_field, module: module, args: [attrs.color, [type: attrs.color_type]]}
+    ]
+
+    root = build_root(root)
+
+    nodes
+    |> Enum.map(&clean_node(&1))
```
polvalente, line 124:
```suggestion
    |> Enum.map(&clean_node/1)
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+     [
+       acc,
+       {{:., [], [{:__aliases__, [alias: false], [module]}, function]}, [], args}
+     ]}
+  end
+
+  defp clean_args(args) do
+    opts =
+      args
+      |> Enum.filter(&is_list(&1))
```
polvalente, line 163:
```suggestion
      |> Enum.filter(&is_list/1)
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+      |> List.flatten()
+      |> Enum.filter(&elem(&1, 1))
+
+    args =
+      args
+      |> Enum.reject(&is_list(&1))
+      |> Enum.reject(&is_nil(&1))
```
polvalente, line 170:
```suggestion
    args = Enum.reject(args, & is_list(&1) or is_nil(&1))
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+
+  function handleFieldChange(event) {
+    const { name, value } = event.target;
+    if (hasType.includes(name)) { maybeReset({name, value}) }
+    ctx.pushEvent("update_field", { field: name, value });
+  }
+
+  ctx.handleEvent("update", ({ fields }) => {
+    setValues(fields);
+  });
```
polvalente, line 112:
```suggestion
  ctx.handleEvent("update", ({ fields }) => setValues(fields));
```

can't these be shortened to one liners like so?

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+      "color" => attrs["color"],
+      "color_type" => attrs["color_type"]
+    }
+
+    {:ok, assign(ctx, fields: fields, options: %{}, vl_alias: nil, missing_dep: missing_dep())}
+  end
+
+  @impl true
+  def scan_binding(pid, binding, env) do
+    send(pid, {:scan_binding_result, binding, env})
```
jonatanklosko, line 28: From the (unpublished) docs:

> **Important:** remember that data sent between processes is copied,
> so avoid sending large data structures. In particular, when looking
> at variables, instead of sending their values, extract and send
> only the relevant metadata.

So basically we want to move the work from `handle_info` here and then `send(pid, {:scan_binding_result, data_options, vl_alias})`, this way we send the minimal amount of data between processes :)

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+    for (const field in fields) {
+      ctx.root.querySelector(`[name="${field}"]`).value = fields[field];
+    }
+  }
+
+  function setDataOptions(options) {
+    const selectElem = ctx.root.querySelector(`[name="data"]`);
+    selectElem.innerHTML = "";
+    createOptions(selectElem, Object.keys(options))
+    const selected = selectElem.value;
```
jonatanklosko, line 136: There's a catch! Let's say we have a variable with data and the chart builder below using this variable. Now when we re-open the notebook (or restart the runtime and install the deps), the smart cell would start, but it would receive an empty binding, until the above code is executed. In such case, we want to preserve the previously selected variable name (that we persist via `attrs`) until the user changes it, regardless if it's currently available. This is important when there are more options, because we don't want to change it implicitly. It's analogous to the connection variable in #105, here's an example:

https://user-images.githubusercontent.com/17034772/158814743-f5fe9837-b701-4900-b899-5627484f8271.mp4

We can handle it the same way, in particular see the `function renderConnectionSelect` function.

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+    %{field => String.to_integer(value)}
+  end
+
+  defp to_updates(_fields, field, value) when field in @as_atom do
+    %{field => String.to_atom(value)}
+  end
+
+  defp to_updates(_fields, field, value), do: %{field => value}
+
+  defp vl_alias(%Macro.Env{aliases: [{vl_alias, VegaLite}]}), do: vl_alias
```
jonatanklosko, line 82: It may not be the only alias, so we should do `Enum.find_value` instead :)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+    |> Map.put("vl_alias", ctx.assigns.vl_alias)
+  end
+
+  @impl true
+  def to_source(attrs) do
+    attrs
+    |> to_quoted()
+    |> Macro.to_string()
+    |> Code.format_string!()
+    |> IO.iodata_to_binary()
```
jonatanklosko, line 97: Once #105 is merged

```suggestion
    attrs |> to_quoted() |> Kino.Utils.Code.quoted_to_string()
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
+defmodule Kino.SmartCell.ChartBuilder do
+  use Kino.JS, assets_path: "lib/assets/chart_builder"
+  use Kino.JS.Live
+  use Kino.SmartCell, name: "Chart builder"
+
+  @as_int ["width", "height"]
+  @as_atom ["data", "chart", "x_axis_type", "y_axis_type", "color_type"]
```
jonatanklosko, line 7: Nit: it's better to keep those as strings and convert to atoms only when generating the code. This way we keep the same type as serialized/deserialized in `attrs` and there's gonna be less surprises :)

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+      </div>
+    </div>
+  `;
+
+  updateInfoBox(info.missing_dep);
+  setDataOptions(info.options);
+  setValues(info.fields);
+
+  const hasType = ["x_axis", "y_axis", "color"];
+  const types = ["x_axis_type", "y_axis_type", "color_type"];
```
jonatanklosko, line 99:
```suggestion
  const channelFields = ["x_axis", "y_axis", "color"];
  const typeFields = ["x_axis_type", "y_axis_type", "color_type"];
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+      <div>
+        <div class="row">
+          <div class="box">
+            <div class="field">
+              <label class="input-label">Chart</label>
+              <select class="input input--sm" name="chart">
+                <option value="bar">Bar</option>
+                <option value="line">Line</option>
+                <option value="area">Area</option>
+                <option value="point">Point</option>
```
jonatanklosko, line 18: We want to order from the most common type, but I'm not sure how to establish this. @josevalim any opinion?

```suggestion
                <option value="point">Point</option>
                <option value="line">Line</option>
                <option value="bar">Bar</option>
                <option value="area">Area</option>
```

josevalim: No idea. Why not alphabetically? :)

jonatanklosko: In this particular case alphabetically doesn't seem right, because `Area` is probably the rarest of these four? :smile:

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  defp apply_node({field, %{name: function, module: module, args: args}}, acc) do
+    ctx = [context: Elixir, import: Kernel]
+    args = if function == :encode_field, do: List.insert_at(args, 0, field), else: args
+
+    {:|>, ctx,
+     [
+       acc,
+       {{:., [], [{:__aliases__, [alias: false], [module]}, function]}, [], args}
+     ]}
```
jonatanklosko, line 157: I think it can be something like this:

```suggestion
    args = if function == :encode_field, do: List.insert_at(args, 0, field), else: args
    
    quote do
      unquote(acc) |> unquote(module).unquote(function)(unquote_splicing(args))
    end
```

We could likely simplify other places with `quote` too, but it's gonna be easier to refactor when we have some tests for the code generation :) You can test the `to_source` function directly as we do [here](https://github.com/livebook-dev/kino/blob/ead0a6f42727eeec95c450b3bb5632c6ea9133e2/test/kino/smart_cell/sql_test.exs#L63-L173).

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+
+      my_data = %{
+        a: [89, 124, 09, 67, 45],
+        b: [12, 45, 67, 83, 32]
+      }
+
+  Or using Explorer:
+      iris = Explorer.Datasets.iris() |> Explorer.DataFrame.to_map()
+
+  All datasets declared above a chart builder will be available on it.`;
```
josevalim, line 220:
```suggestion
The dataset needs to be a map of series, for example:

      my_data = %{
        a: [89, 124, 09, 67, 45],
        b: [12, 45, 67, 83, 32]
      }

Or using Explorer:

      iris = Explorer.Datasets.iris() |> Explorer.DataFrame.to_map()

Chart builder is available for all datasets above.`;
```

jonatanklosko: > Chart builder is available for all datasets above

I'm wondering if it makes sense to say this, given it's the core principle in Livebook in general.

lib/assets/chart_builder/main.js
cristineguadelupe: I feel that Bar and Line are the most common/familiar (especially for beginners) so my suggestion is Bar, Line, Point, Area. Wdy?

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,223 @@
...
+  function resetField(elem) {
+    elem.value = "";
+    elem.disabled = true;
+    ctx.pushEvent("update_field", { field: elem.name, value: "" });
+  }
+
+  function createOptions(selectElem, options) {
+    for (const option of options) {
+      const element = document.createElement("option");
+      element.innerText = option;
```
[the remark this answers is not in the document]
cristineguadelupe: This is cool but I need more help here! I ended up with more code for the elements where I need to append or prepend options. Or with two different ways to create the options depending on the rules.

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,289 @@
...
+
+  updateInfoBox(info.missing_dep);
+  info.fresh ? setFreshData(info.options) : renderCurrentData(info.fields);
+
+  ctx.root.addEventListener("blur", handleFieldChange, true);
+  ctx.root.addEventListener("change", handleFieldChange);
+
+  function handleFieldChange(event) {
+    const { name, value } = event.target;
+    if (channelFields.includes(name)) { maybeReset({name, value}) };
```
jonatanklosko, line 105:
```suggestion
    if (channelFields.includes(name)) { maybeReset(name, value) };
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+    setTypes();
+  }
+
+  function setTypes() {
+    for (const elem of typeFields) {
+      const selectElem = ctx.root.querySelector(`[name="${elem}"]`);
+      selectElem.selectedIndex = 0;
+      const selected = selectElem.value;
+      elem === "color_type" ? selectElem.disabled = true : selectElem.disabled = false;
```
jonatanklosko, line 190:
```suggestion
      selectElem.disabled = (elem === "color_type");
```

lib/assets/chart_builder/main.js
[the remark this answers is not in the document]
jonatanklosko: I'd say it's enough to make the select dimmed, rather than individual options (that's what we do for the SQL cell). Then you could conditionally prepend an option to the list and render all of them. That said, I'm ok with either approach, and creating elements directly is fine, so your call :)

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,289 @@
...
+    updateDataInfoBox(fields["data"]);
+  }
+
+  function setFreshData(options) {
+    enableAll();
+    const selectElem = ctx.root.querySelector(`[name="data"]`);
+    selectElem.innerHTML = "";
+    createOptions(selectElem, Object.keys(options));
+    const selected = selectElem.value;
+    updateDataInfoBox(options);
```
jonatanklosko, line 168: The data info message no longer shows, this should probably be

```suggestion
    updateDataInfoBox(options[selected]);
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,206 @@
...
+
+    quote do
+      unquote(acc) |> unquote(module).unquote(function)(unquote_splicing(args))
+    end
+  end
+
+  defp clean_args(args) do
+    opts =
+      args
+      |> Enum.filter(&is_list/1)
```
jonatanklosko, line 192: I think you can simplify if you make `:opts` a separate key in the node, since we know what any options should go as the last argument :)

jonatanklosko (commented):
Awesome, a couple more comments from me :)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,206 @@
...
+  defp convert_field(field, value) when field in @as_atom do
+    {String.to_atom(field), String.to_atom(value)}
+  end
+
+  defp convert_field(field, value), do: {String.to_atom(field), value}
+
+  defp vl_alias(%Macro.Env{aliases: aliases}) do
+    Enum.find_value(aliases, fn {current_alias, module} ->
+      if module == VegaLite, do: current_alias
+    end)
```
josevalim, line 99:
```suggestion
    case List.keyfind(aliases, VegaLite, 1) do
      {alias, _} -> alias
      nil -> nil
    end
```

lib/kino/smart_cell/chart_builder.ex
josevalim: I would propose another approach instead. Instead of having the node do this:

        args: [attrs.color, [type: attrs.color_type]]

I think you should have something like this:

        args: maybe_arg_type(attrs.color, attrs.color_type)

Which builds the correct argument in the first place (or set it to `nil` so it is discarded later).

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,206 @@
...
+      "width" => attrs["width"] || "",
+      "height" => attrs["height"] || "",
+      "x_axis" => attrs["x_axis"] || "",
+      "y_axis" => attrs["y_axis"] || "",
+      "x_axis_type" => attrs["x_axis_type"] || "",
+      "y_axis_type" => attrs["y_axis_type"] || "",
+      "color" => attrs["color"] || "",
+      "color_type" => attrs["color_type"] || "",
+      "data" => attrs["data"] || ""
+    }
```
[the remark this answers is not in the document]
cristineguadelupe: It would be okay by me, but I honestly don't see the benefits. We already have consistency since we have empyt strings for JS and `nil` for Elixir, and the conversion happens only on `to_source`. I also think it would lead us to have comparisons with  `=== "nil"` on JS's side which doesn't hurt but seems a bit odd... Unless I'm missing something here 🤔

lib/assets/chart_builder/main.js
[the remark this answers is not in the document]
cristineguadelupe: I'll keep it as it is for now! :)

lib/kino/smart_cell/chart_builder.ex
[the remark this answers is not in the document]
jonatanklosko: Ah, I wrote that before I realised that `convert_field` doesn't happen on change, but before code generation.

I'm mostly wondering about the types we should have in `attrs` (because that's what we persist and it's hard to change later on). It may be convenient to just keep the values coming from JS, which are pretty much always strings, but I think keeping the parsed values is cleaner. So for fields that are selects either value or `nil`, for numbers either integer or `nil`, for text fields always string. In other words, I'd lean towards parsing the data on change, though I don't have a strong reason other than it _feels_ better to me and I believe that's what we do in the other cells :)

> I also think it would lead us to have comparisons with `=== "nil"`

`nil` becomes `null` on JS side :)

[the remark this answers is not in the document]
cristineguadelupe: Done 🐈

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,205 @@
...
+
+  defp to_quoted(attrs) do
+    attrs = Map.new(attrs, fn {k, v} -> convert_field(k, v) end)
+
+    [root | nodes] = [
+      %{
+        field: nil,
+        name: :new,
+        module: attrs.vl_alias,
+        args: [[width: attrs.width, height: attrs.height]]
```
josevalim, line 138: You could do `build_arg_root(height, width)` here and simplify the build_root later on. :) similar to what we did with `arg_type`!

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,205 @@
...
+
+    {:ok, ctx, reevaluate_on_change: true}
+  end
+
+  @impl true
+  def scan_binding(pid, binding, env) do
+    data_options =
+      for {key, val} <- binding, is_atom(key), is_map(val), into: %{}, do: {key, Map.keys(val)}
+
+    vl_alias = vl_alias(env)
```
jonatanklosko, line 41: When we generate the code with alias and then open the notebook, as soon as the cell starts (the first cell gets evaluated) we would generate a new code without the alias, and then update once the alias is created. I'd say it's fine, @josevalim wdyt?

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,205 @@
...
+    {:noreply, ctx}
+  end
+
+  @impl true
+  def handle_event("update_field", %{"field" => field, "value" => value}, ctx) do
+    current_data = ctx.assigns.fields["data_variable"]
+    current_field = ctx.assigns.fields[field]
+
+    updated_fields = to_update(field, value)
+    ctx = update(ctx, :fields, &Map.merge(&1, updated_fields))
```
jonatanklosko, line 71: Note that we compare the value below, so we want to use the parsed one for that.

```suggestion
    value = parse_value(field, value)
    ctx = update(ctx, :fields, &Map.put(&1, field, value))
```

(in the db connection cell we have `to_updates`, because changing database type also updates the port, here it's always a single update)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,205 @@
...
+
+    if field == "data_variable" && value != current_data, do: update_options(ctx, value)
```
jonatanklosko, line 73: Doesn't the client have all options from the `"set_available_data"` event?

lib/kino/smart_cell/chart_builder.ex
josevalim: @jonatanklosko fine by me.

lib/kino/smart_cell/chart_builder.ex
cristineguadelupe: Done!

cristineguadelupe:
> Should we rather the infobox generic and build the error message from Elixir? This way we write less JavaScript... something like this:
> 
> ```
>   ctx.handleEvent("error", ({ message }) => {
>     updateInfoBox(message);
> ```

Since we are moving more things to JS I’m not sure if that’s make sense for now

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,301 @@
...
+        selectElem.classList.add("unavailable");
+      } else {
+        selectElem.classList.remove("unavailable");
+      }
+    }
+  }
+
+  function maybeReset(name, value) {
+    const elem = ctx.root.querySelector(`[name="${name}_type"]`);
+    !value ? resetField(elem) : elem.disabled = false;
```
jonatanklosko, line 226:
```suggestion
    if (value) {
      elem.disabled = false;
    } else {
      resetField(elem)
    }
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,301 @@
...
+        selectElem.value = selected;
+        selectElem.classList.add("unavailable");
+      } else {
+        selectElem.classList.remove("unavailable");
+      }
+    }
+  }
+
+  function maybeReset(name, value) {
+    const elem = ctx.root.querySelector(`[name="${name}_type"]`);
```
jonatanklosko, line 225:
```suggestion
  function maybeResetChannelType(name, value) {
    const typeElem = ctx.root.querySelector(`[name="${name}_type"]`);
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,301 @@
...
+    for (const field in fields) {
+      ctx.root.querySelector(`[name="${field}"]`).value = fields[field];
+    }
+  }
+
+  function setAvailableData(options) {
+    enableAll();
+    const selectedData = selectDataElem.value;
+    if (!selectedData) {return setFreshData(options)};
+    const availableDataOptions = options[selectedData];
```
jonatanklosko, line 145: ?

```suggestion
    const axisOptions = options[selectedData];
```

lib/assets/chart_builder/main.js
```diff
@@ -0,0 +1,301 @@
...
+    enableAll();
+    for (const elem of channelFields) {
+      const selectElem = ctx.root.querySelector(`[name="${elem}"]`);
+      const inner = elem === "color_field" ? "<option value=''></option>" : "";
+      selectElem.innerHTML = inner;
+      createOptions(selectElem, options);
+      const selected = selectElem.value;
+      ctx.pushEvent("update_field", { field: elem, value: selected });
+    }
+    setTypes();
```
jonatanklosko, line 193: We are sending `"update_field"` in the loop above, and then in `setTypes` in a loop as well. We shouldn't be spamming the server so much.

This bulk change is a result of selecting the dataframe, to set the new defaults we can either:

1. Compute the new values on the client and send as multi-field update, then broadcast all those fields
2. Send just the new dataframe name and determine the new values on the server, then broadcast all those fields

I think 2. is more direct and will end up better :)

jonatanklosko (commented):
Looks great, just one more important detail and we should be good :)

josevalim (approved):
All good from my side. The JS is up to Jonatan :D

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,210 @@
...
+    ctx = update(ctx, :fields, &Map.put(&1, field, parsed_value))
+    broadcast_event(ctx, "update", %{"fields" => %{field => parsed_value}})
+
+    {:noreply, ctx}
+  end
+
+  defp update_data_options(ctx, value) do
+    df = String.to_atom(value)
+    axis = ctx.assigns.options[df] |> List.first()
+    axis = if axis, do: Atom.to_string(axis)
```
jonatanklosko, line 89: Would it make more sense to default x and y to different columns?

```suggestion
    {x_field, y_field} =
      case ctx.assigns.options[df] do
        [] -> {nil, nil}
        [key] -> {Atom.to_string(key), Atom.to_string(key)}
        [key1, key2 | _] -> {Atom.to_string(key1), Atom.to_string(key2)}
      end
```

Also, we could convert the column names to strings upfront in `scan_binding`, since we don't need them as atoms.

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,210 @@
...
+  @impl true
+  def handle_info({:scan_binding_result, data_options, vl_alias}, ctx) do
+    ctx = assign(ctx, options: data_options, vl_alias: vl_alias)
+    broadcast_event(ctx, "set_available_data", %{"options" => data_options})
+
+    {:noreply, ctx}
+  end
+
+  @impl true
+  def handle_event("update_field", %{"field" => "data_variable", "value" => ""}, ctx) do
```
jonatanklosko, line 66: Shouldn't we reset other fields in this case? We can match in `update_data_options` on empty value, and return all empty fields :)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,210 @@
...
+
+  def handle_event("update_field", %{"field" => field, "value" => value}, ctx) do
+    parsed_value = parse_value(field, value)
+    ctx = update(ctx, :fields, &Map.put(&1, field, parsed_value))
+    broadcast_event(ctx, "update", %{"fields" => %{field => parsed_value}})
+
+    {:noreply, ctx}
+  end
+
+  defp update_data_options(ctx, value) do
```
jonatanklosko, line 86: `update_` suggests this function changes `ctx` or has a side effect, maybe `updates_for_data_variable`?

jonatanklosko (commented):
Final nitpicks and ship it! :)

lib/kino/smart_cell/chart_builder.ex
cristineguadelupe: Nops! Here we have a specific case. `data_variable` will be `""` if and only if the ChartBuilder has never been populated before. So we have nothing to reset and we don't need to broadcast.

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,225 @@
...
+  end
+
+  @impl true
+  def scan_binding(pid, binding, env) do
+    data_options =
+      for {key, val} <- binding,
+          is_atom(key),
+          is_map(val),
+          into: %{},
+          do: {Atom.to_string(key), Map.keys(val)}
```
jonatanklosko, line 43:
```suggestion
          do: {Atom.to_string(key), val |> Map.keys() |> Enum.map(&to_string/1)}
```

lib/kino/smart_cell/chart_builder.ex
cristineguadelupe: This happens in two scenarios: when a ChartBuilder is started without variables and when reconnection with an empty (or fresh) ChartBuilder occurs.

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -0,0 +1,225 @@
...
+    broadcast_event(ctx, "update", %{"fields" => %{field => parsed_value}})
+
+    {:noreply, ctx}
+  end
+
+  defp updates_for_data_variable(ctx, value) do
+    {x_field, y_field} =
+      case ctx.assigns.options[value] do
+        [key] -> {Atom.to_string(key), Atom.to_string(key)}
+        [key1, key2 | _] -> {Atom.to_string(key1), Atom.to_string(key2)}
```
jonatanklosko, line 94: Actually the map keys may as well be strings or numbers, so we probably want `to_string` instead, but we can as well do it when scanning too.

```suggestion
        [key] -> {key, key}
        [key1, key2 | _] -> {key1, key2}
```

lib/kino/smart_cell/chart_builder.ex
jonatanklosko: What I meant is that if it's the JS side triggering and we don't want it, we can always add a check like

```js
if (name === "data_variable" && value === "") { return }
```

lib/kino/smart_cell/chart_builder.ex
jonatanklosko: Ah, in fact the map may have keys that we cannot convert to strings, we should filter only those that have atom/string keys then.

(in the future we will have a separate abstraction for tabular data that will be more precise and flexible, so for now we just don't want to crash ^^)

cristineguadelupe: `data = %{3 => [1, 2, 3], 4 => [1, 2, 3]}` this is working! Are there other cases?

jonatanklosko: The problem is with terms that don't support `to_string`, like this:

```
data = %{self() => [1, 2, 3]}
```

cristineguadelupe: I see! I tried to solve it for now. Thanks!

jonatanklosko (approved):
Fantastic :cat:

PR #108, opened 2022-03-21
Generate unique default names for smart cell variables

jonatanklosko:
https://user-images.githubusercontent.com/17034772/159354243-ca904a8c-d549-4e84-980d-b60b55ff86c8.mp4

### DIFF
--- a/lib/kino/counter.ex
+++ b/lib/kino/counter.ex
@@ -1,0 +1,35 @@
+defmodule Kino.Counter do
+  @moduledoc false
+
+  @table_name __MODULE__
+
+  @doc """
+  Initializes resources for global counters.
+  """
+  @spec initialize() :: :ok
+  def initialize() do
+    :ets.new(@table_name, [:set, :named_table, :public])
+    :ok
+  end
+
+  @doc """
+  Increments the given counter and returns the new value.
+  """
+  @spec next(term()) :: integer()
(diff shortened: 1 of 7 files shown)


lib/kino/counter.ex
```diff
@@ -0,0 +1,35 @@
...
+    :ets.update_counter(@table_name, key, {2, 1}, {key, 0})
+  end
+
+  @doc """
+  Bumps the counter to `value` unless it already has a higher value.
+
+  Returns the current counter value.
+  """
+  @spec bump(term(), integer()) :: integer()
+  def bump(key, value) do
```
josevalim, line 29: `bump` typically means "bump by value". Maybe set or put would be better here? :)

josevalim (approved):
Beautiful, just one nitpick!

lib/kino/counter.ex
jonatanklosko: `set` and `put` sound like they would always override the value, while in this case we only do it if it increases :thinking:

josevalim: @jonatanklosko `at_least`, `update_max`, `put_max`, all work. Your call. :)

josevalim: I mean they all work _in addition_ to what you suggested. :)

PR #110, opened 2022-03-23
Keep initial chart builder fields enabled when the data is already available

jonatanklosko:
I restarted the runtime and then evaluated everything using `Ctrl+Shift+Enter`. In such case, by the time the smart cell loads, everything is evaluated, so it receives the available data in the initial payload, so we don't want to disable the fields.

### DIFF
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -95,17 +95,22 @@ export function init(ctx, payload) {
   const typeFields = ["x_field_type", "y_field_type", "color_field_type"];
 
   const state = {
-    data_options: payload.options,
+    dataOptions: payload.data_options,
   };
 
   const selectDataElem = ctx.root.querySelector(`[name="data_variable"]`);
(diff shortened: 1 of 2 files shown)


lib/assets/chart_builder/main.js
```diff
@@ -166,8 +171,12 @@ export function init(ctx, payload) {
     }
   }
 
-  function renderCurrentData(fields) {
-    disableAll();
+  function renderCurrentData(fields, dataOptions) {
+    const dataVariable = fields["data_variable"];
+    if (!dataOptions[dataVariable]) {
+      disableAll();
+    }
```
jonatanklosko, line 178: This is the relevant part.

The remaining diff renaming `options` to explicitly be either data or axis options, as otherwise it's quite confusing which is which :cat2:

cristineguadelupe: Thanks 😊

cristineguadelupe (approved):

PR #115, opened 2022-03-31
Count as special option

cristineguadelupe:
Moves count aggregation to the field options

### DIFF
--- a/lib/kino/smart_cell/chart_builder.ex
+++ b/lib/kino/smart_cell/chart_builder.ex
@@ -175,19 +175,19 @@ defmodule Kino.SmartCell.ChartBuilder do
       %{field: :mark, name: :mark, module: attrs.vl_alias, args: [attrs.chart_type]},
       %{
         field: :x,
-        name: :encode_field,
+        name: encode(attrs.x_field),
         module: attrs.vl_alias,
         args: build_arg_field(attrs.x_field, attrs.x_field_type, attrs.x_field_aggregate)
       },
       %{
         field: :y,
-        name: :encode_field,
+        name: encode(attrs.y_field),
         module: attrs.vl_alias,
         args: build_arg_field(attrs.y_field, attrs.y_field_type, attrs.y_field_aggregate)
       },
       %{
         field: :color,
-        name: :encode_field,
+        name: encode(attrs.color_field),
         module: attrs.vl_alias,
         args:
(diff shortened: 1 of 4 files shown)


josevalim:
Nitpick: to avoid confusion with field names, what if we call it `COUNT` or even `COUNT(*)` if we want to mirror deepnote?

cristineguadelupe:
> Nitpick: to avoid confusion with field names, what if we call it `COUNT` or even `COUNT(*)` if we want to mirror deepnote?

It's already in bold. Do you think that's not enough?

josevalim:
Oh, I didn't notice it. So let's go with `COUNT` in bold, that should be enough!

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -207,7 +207,10 @@ defmodule Kino.SmartCell.ChartBuilder do
   defp apply_node(%{args: nil}, acc), do: acc
 
   defp apply_node(%{field: field, name: function, module: module, args: args}, acc) do
-    args = if function == :encode_field, do: List.insert_at(args, 0, field), else: args
+    args =
+      if function == :encode_field || function == :encode,
+        do: List.insert_at(args, 0, field),
+        else: args
```
jonatanklosko, line 213:
```suggestion
    args = if function in [:encode_field, :encode], do: [field | args], else: args
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -224,6 +227,7 @@ defmodule Kino.SmartCell.ChartBuilder do
   end
 
   defp build_arg_field(nil, _, _), do: nil
+  defp build_arg_field("count", _, _), do: [[aggregate: :count]]
```
jonatanklosko, line 230: I can imagine `"count"` being a regular dataframe column, so for the value let's use something like `"__count__"`.

For the same reason it would make sense to make the label more distinctive. I actually like `COUNT(*)` the most because it matches SQL, but `COUNT` in all capitals is also alright.

jonatanklosko:
I'd also keep `COUNT(*)` bold-less, so they don't think it's selected (takes too much attention imo), but your call :)

jonatanklosko (approved):

josevalim:
Bold-less sounds good to me too, otherwise you are right it calls too much attention. Then ship it!

PR #116, opened 2022-04-01
Fix: removes count from data variable

cristineguadelupe:
🙈

### DIFF
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -286,7 +286,9 @@ export function init(ctx, payload) {
       if (elemClass) { element.classList.add(elemClass) };
       selectElem.append(element);
     }
-    if (elemClass !== "unavailable-option") {appendCountOption(selectElem);}
+    if (elemClass !== "unavailable-option" && selectElem.name !== "data_variable") {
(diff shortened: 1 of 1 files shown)


lib/assets/chart_builder/main.js
```diff
@@ -286,7 +286,9 @@ export function init(ctx, payload) {
       if (elemClass) { element.classList.add(elemClass) };
       selectElem.append(element);
     }
-    if (elemClass !== "unavailable-option") {appendCountOption(selectElem);}
+    if (elemClass !== "unavailable-option" && selectElem.name !== "data_variable") {
```
jonatanklosko, line 289: Just guessing

```suggestion
    if (channelFields.includes(selectElem.name)) {
```

jonatanklosko: Though I think we still need the unavailable option check

jonatanklosko (approved):

lib/assets/chart_builder/main.js
cristineguadelupe: Yep! We need it for now. And again, Vue is coming to save us. ;)

PR #117, opened 2022-04-04
Refactor to Vue

cristineguadelupe:
The only small change in behavior is that I’m no longer resetting the type fields at each channel field change. This turned out to be a not so good user experience. I feel we have a more consistent behavior between all fields now.

### DIFF
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -166,13 +166,6 @@ export function init(ctx, payload) {
       };
     },
 
-    mounted() {
-      if (payload.fresh && this.dataVariables.length) {
-        this.fields.data_variable = Object.keys(this.dataOptions)[0];
-        ctx.pushEvent("update_field", { field: "data_variable", value: this.fields.data_variable });
-      }
-    },
-
     computed: {
       axisOptions() {
         const data = this.fields.data_variable;
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -77,7 +77,7 @@ export function init(ctx, payload) {
             <div class="group">
               <BaseSelect
                 name="x_field"
-                label="X Axis"
+                label="X axis"
                 v-model="fields.x_field"
                 :options="axisOptions"
                 :required
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -279,9 +279,10 @@ export function init(ctx, payload) {
     app.missingDep = dep;
   });
 
-  ctx.handleEvent("set_available_data", ({ data_options }) => {
+  ctx.handleEvent("set_available_data", ({ data_options, fields }) => {
     app.dataVariables = data_options.map(data => data["variable"]);
     app.dataOptions = data_options;
+    setValues(fields);
   });
 
   function setValues(fields) {
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -81,7 +81,6 @@ export function init(ctx, payload) {
                 v-model="fields.x_field"
                 :options="axisOptions"
                 :required
-                :count
                 :disabled="noDataVariable"
               />
               <BaseSelect
@@ -106,7 +105,6 @@ export function init(ctx, payload) {
                 v-model="fields.y_field"
                 :options="axisOptions"
                 :required
-                :count
                 :disabled="noDataVariable"
               />
               <BaseSelect
@@ -130,7 +128,6 @@ export function init(ctx, payload) {
                 label="Color"
                 v-model="fields.color_field"
                 :options="axisOptions"
-                :count
                 :disabled="noDataVariable"
               />
               <BaseSelect
@@ -170,7 +167,7 @@ export function init(ctx, payload) {
       axisOptions() {
         const dataVariable = this.fields.data_variable;
         const dataOptions = this.dataOptions.find(data => data["variable"] === dataVariable);
-        return dataOptions ? dataOptions["columns"] : [];
+        return dataOptions ? dataOptions["columns"].concat("__count__") : [];
       },
       noDataVariable() {
         return !this.fields.data_variable;
@@ -236,18 +233,13 @@ export function init(ctx, payload) {
             type: Boolean,
             default: false
           },
-          count: {
-            type: Boolean,
-            default: false
-          },
         },
         methods: {
           available(value, options) {
-            if (value === "__count__" || !value) {
-              return true;
-            } else {
-              return options.includes(value);
-            }
+            return value ? options.includes(value) : true;
+          },
+          optionLabel(value) {
+            return value === "__count__" ? "COUNT(*)" : value;
           },
         },
         template: `
@@ -260,18 +252,18 @@ export function init(ctx, payload) {
               class="input input--sm"
               :class="{ unavailable: !available(modelValue, options) }"
             >
-              <option v-if="!required"></option>
+              <option v-if="!required && available(modelValue, options)"></option>
               <option
                 v-for="option in options"
                 :value="option"
                 :key="option"
                 :selected="option === modelValue"
-              >{{ option }}</option>
-              <option v-if="count" value="__count__">COUNT(*)</option>
+              >{{ optionLabel(option) }}</option>
               <option
                 v-if="!available(modelValue, options)"
                 class="unavailable-option"
-              >{{ modelValue }}</option>
+                :value="modelValue"
+              >{{ optionLabel(modelValue) }}</option>
             </select>
           </div>
         `
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -179,7 +179,7 @@ export function init(ctx, payload) {
         return this.dataOptions[data];
       },
       noDataVariable() {
-        return !this.fields.data_variable
+        return !this.fields.data_variable;
       },
       noColorField() {
         return !this.fields.color_field || this.fields.color_field === "__count__";
--- a/lib/assets/chart_builder/main.js
+++ b/lib/assets/chart_builder/main.js
@@ -1,358 +1,306 @@
+import * as Vue from "https://cdn.jsdelivr.net/npm/vue@3.2.26/dist/vue.esm-browser.prod.js";
+
 export function init(ctx, payload) {
   ctx.importCSS("main.css");
   ctx.importCSS("https://fonts.googleapis.com/css2?family=Inter:wght@400;500&display=swap");
 
-  ctx.root.innerHTML = `
-    <div class="app">
-      <div id="info-box" class="info-box"></div>
-      <div id="data-info-box" class="info-box"></div>
-      <div>
-        <div class="row">
-          <div class="box">
-            <div class="field">
-              <label class="input-label">Data</label>
-              <select class="input input--sm" name="data_variable">
-              </select>
+  const app = Vue.createApp({
+    template: `
+      <div class="app">
+        <!-- Info Messages -->
+        <div id="info-box" class="info-box" v-if="missingDep">
+          <p>To successfully build charts, you need to add the following dependency:</p>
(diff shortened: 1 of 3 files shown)


lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
-            <div class="field">
-              <label class="input-label">Title</label>
-              <input class="input input--md" name="chart_title" />
-            </div>
-          </div>
+  const app = Vue.createApp({
+    template: `
+      <div class="app">
+        <!-- Info Messages -->
+        <div id="info-box" class="info-box" v-if="missingDep">
```
philss, line 11: I don't remember much of Vue, but does the `id` stays global? Or does it only exist for the scope of this app? If is the former, then I would suggest to have a more "scoped" id, like `chart-builder-info-box` or something like that. WDYT?

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
-            </div>
-          </div>
-          <div class="group">
-            <div class="field">
-              <label class="input-label">X Axis</label>
-              <select class="input input--sm" name="x_field">
-              </select>
+        <div id="data-info-box" class="info-box" v-if="noDataVariable">
+          <p>To successfully plot graphs, you need at least one dataset available.</p>
+          <p>The dataset needs to be a map of series, for exemple:</p>
```
philss, line 17:
```suggestion
          <p>The dataset needs to be a map of series, for example:</p>
```

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
+      }
+    },
+
+    computed: {
+      axisOptions() {
+        const data = this.fields.data_variable;
+        return this.dataOptions[data];
+      },
+      noDataVariable() {
+        return !this.fields.data_variable
```
philss, line 182:
```suggestion
        return !this.fields.data_variable;
```
For consistency :)

philss (commented):
Cool! I added some minor points, but LGTM! :ship:

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
-              <label class="input-label">Aggregate</label>
-              <select class="input input--sm" name="y_field_aggregate">
-                <option value=""></option>
-                <option value="mean">Mean</option>
-                <option value="sum">Sum</option>
-              </select>
+            <div class="group">
+              <BaseSelect
+                name="x_field"
+                label="X Axis"
```
jonatanklosko, line 80:
```suggestion
                label="X axis"
```

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
-          </div>
-          <div class="group">
-            <div class="field">
-              <label class="input-label">Color</label>
-              <select class="input input--sm" name="color_field">
-              </select>
+            <div class="group">
+              <BaseSelect
+                name="y_field"
+                label="Y Field"
```
jonatanklosko, line 105:
```suggestion
                label="Y axis"
```

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
+        typeOptions: ["quantitative", "nominal", "ordinal"],
+        aggregateOptions: ["sum", "mean"],
+        dataVariables: Object.keys(payload.data_options),
+      };
+    },
+
+    mounted() {
+      if (payload.fresh && this.dataVariables.length) {
+        this.fields.data_variable = Object.keys(this.dataOptions)[0];
+        ctx.pushEvent("update_field", { field: "data_variable", value: this.fields.data_variable });
```
jonatanklosko, line 172: Oh, we should have this logic on the server! When we scan binding and discover data variables, but the current `data_variable` is `nil`, we should set it to the first one and broadcast together with the new variables. Having that you probably no longer need the `fresh` check :)

jonatanklosko (commented):
:cat:

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,306 @@
...
+              :class="{ unavailable: !available(modelValue, options) }"
+            >
+              <option v-if="!required"></option>
+              <option
+                v-for="option in options"
+                :value="option"
+                :key="option"
+                :selected="option === modelValue"
+              >{{ option }}</option>
+              <option v-if="count" value="__count__">COUNT(*)</option>
```
jonatanklosko, line 276: We need to replicate `:selected` here too. One option to simplify would be to include `__count__` in `axisOptions()`. Then for the label we can either have mapping, or maybe just make the options a list of `[key, label]` altogether, whatever is simpler :)

cristineguadelupe: Actually we don't need to because this option is outside of v-for and will be auto-selected. :)
I quite like the solution we have today. We don't have to manipulate the data frame options and add something that doesn't really belong there.
We don't have to worry about conditionals to know which fields should or should not add the count. Today the field that wants to support the count just needs to make it explicit and we will do the rest!

lib/assets/chart_builder/main.js
jonatanklosko: > and the scan binding doing too much

Not that the check is expensive, but we would do it in `handle_info`, so it's fine either way

I don't think it's that complicated, we already do something analogous here:

https://github.com/livebook-dev/kino/blob/c6194c743299ada79d94ff12a0fd678a0aa53dfd/lib/kino/smart_cell/sql.ex#L71-L74

(connections -> data variables, connection -> current data variable)

jonatanklosko: Btw, my primary concern with the current version is that it doesn't feel right to send a field update to the server when the client just `mount`s :)

lib/assets/chart_builder/main.js
jonatanklosko: It feels a bit off to have a generic `BaseSelect` and then an axis-specific `__count__` option there. I was mostly wondering how we could minimize knowing about `__count__` everywhere.

> We don't have to worry about conditionals to know which fields should or should not add the count.

In the alternative version we wouldn't either, since we have `axisOptions()`, we can just include `__count__` there.

That said, it's up to you, I'm mostly throwing some ideas :)

cristineguadelupe: Why do you think it’s a bit off?
The repetition is not because count is not on the array, but because count behaves differently. So I need to check anyway.
Maybe we can use a const to storage it’s value…  How about?

lib/assets/chart_builder/main.js
cristineguadelupe: Which check is expensive?
Well, in this case we are actually doing an update for the user. It’s in fact a field update! If we change it to the server side, we will need to broadcast it anyway. Vue needs to know that’s a change to trigger everything correctly.
Do you have some code suggestions u think solves it?
My main concern is about the unavailable axis options. Just broadcast the set_available_data and the updated fields won’t work.
I guess we will need a specific broadcast for this case to force vue to trigger everything manually.

jonatanklosko: Why can't we send available data together with the selected data variable and do:

```
...
app.dataOptions = data_options;
app.fields.data_variable = data_variable;
```

That's what would happen if we received a regular field update, no?

lib/assets/chart_builder/main.js
jonatanklosko: All I mean is that count is a special option specific to axis, so if we have it among the options passed to the component, then the component doesn't need to know about it.

> Maybe we can use a const to storage it’s value… How about?

Do you mean the label? That was my main question, if the label should be `{LABELS[option] || option}` (which is rather messy), or if rather the options should be `[key, label]` pairs, which I guess may not be that convenient here. So as said, I'm fine keeping it as is, we can always revisit if there are more special options :)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -74,9 +72,20 @@ defmodule Kino.SmartCell.ChartBuilder do
...
+      ctx = update(ctx, :fields, &Map.merge(&1, updated_fields))
+      broadcast_event(ctx, "update", %{"fields" => updated_fields})
+      broadcast_event(ctx, "set_available_data", %{"data_options" => data_options})
+
+      {:noreply, ctx}
+    else
+      broadcast_event(ctx, "set_available_data", %{"data_options" => data_options})
+
+      {:noreply, ctx}
+    end
```
jonatanklosko, line 88: We can broadcast just once, like this:

```suggestion
    updated_fields =
      case {ctx.assigns.fields["data_variable"], data_options} do
        {nil, [{data_variable, _} | _]} -> updates_for_data_variable(ctx, data_variable)
        _ -> %{}
      end

    ctx = update(ctx, :fields, &Map.merge(&1, updated_fields))

    broadcast_event(ctx, "set_available_data", %{
      "data_options" => data_options,
      "fields" => updated_fields
    })
```

And on the JS side we just need to do `setValues(fields);` in the handler, as would happen with the separate event.

Note: the code above assumes that `data_options` is a list of pairs, instead of a map, and we should make that change. Just recently we changed Livebook so that variables are ordered from latest to oldest, so we need a list to keep the order :)

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -74,6 +74,14 @@ defmodule Kino.SmartCell.ChartBuilder do
...
     ctx = assign(ctx, data_options: data_options, vl_alias: vl_alias)
+
+    updated_fields =
+      case {ctx.assigns.fields["data_variable"], data_options} do
+        {nil, [%{variable: data_variable} | _]} -> updates_for_data_variable(ctx, data_variable)
+        _ -> %{}
+      end
+
+    ctx = update(ctx, :fields, &Map.merge(&1, updated_fields))
     broadcast_event(ctx, "set_available_data", %{"data_options" => data_options})
```
jonatanklosko, line 85: We still need to send the fields and `setValues(fields)` on the client, right?

```suggestion
    broadcast_event(ctx, "set_available_data", %{"data_options" => data_options, "fields" => updated_fields})
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -97,8 +105,16 @@ defmodule Kino.SmartCell.ChartBuilder do
   end
 
   defp updates_for_data_variable(ctx, value) do
+    options =
+      ctx.assigns.data_options
+      |> Enum.find(&(&1.variable == value))
+      |> case do
+        nil -> []
+        options -> options.columns
+      end
```
jonatanklosko, line 114: :cat2:

```suggestion
    columns =
      Enum.find_value(ctx.assigns.data_options, [], &(&1.variable == value && &1.columns))
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -97,8 +105,16 @@ defmodule Kino.SmartCell.ChartBuilder do
...
+
     {x_field, y_field} =
-      case ctx.assigns.data_options[value] do
+      case options do
```
jonatanklosko, line 117:
```suggestion
      case columns do
```

lib/assets/chart_builder/main.js
```diff
@@ -1,358 +1,292 @@
...
+
+    data() {
+      return {
+        fields: payload.fields,
+        dataOptions: payload.data_options,
+        missingDep: payload.missing_dep,
+        chartOptions: ["bar", "line", "point", "area"],
+        typeOptions: ["quantitative", "nominal", "ordinal"],
+        aggregateOptions: ["sum", "mean"],
+        dataVariables: payload.data_options.map(data => data["variable"]),
```
jonatanklosko, line 162:
```suggestion
        dataVariables: payload.data_options.map(data => data.variable),
```

(either is fine, just saying this is the same :))

jonatanklosko (commented):
Awesome, this is much cleaner now! A couple final comments and ship it :D

jonatanklosko (approved):
:shipit:

PR #120, opened 2022-04-06
Chart builder - responsiveness

cristineguadelupe:
CleanShot 2022-04-06 at 20 08 00@2x
CleanShot 2022-04-06 at 20 09 59@2x

### DIFF
--- a/lib/assets/chart_builder/main.css
+++ b/lib/assets/chart_builder/main.css
@@ -14,7 +14,10 @@
 
 .row {
   display: flex;
-  align-items: center;
+  flex-direction: row;
+  align-items: flex-start;
+  align-content: stretch;
+  flex-wrap: nowrap;
 }
 
 .row > *:not(:first-child) {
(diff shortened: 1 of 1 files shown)


lib/assets/chart_builder/main.css
```diff
@@ -25,29 +28,39 @@
...
 }
 
-.box {
+.box:nth-child(3) {
+  flex: 0 1 54%;
+}
+
+.box:nth-child(2) .field:not(:first-child) {
+  padding-left: 12px;
+}
```
jonatanklosko, line 49: These focus specific elements, right? If so, it may be worth adding more classes to reduce mental overhead :)

jonatanklosko (approved):
Just one comment, feel free to merge when you are happy :)

lib/assets/chart_builder/main.css
cristineguadelupe: Sure thing! I’m just doing the bare minimum until we have the final design. That’s the reason why I’ve decided not to create any more classes for now.

jonatanklosko: Ah sure, sound good to me!

PR #122, opened 2022-04-08
Convert DataTable to use table reader

jonatanklosko:
Changes `DataTable` to accept tabular data. It still supports a list of structs and we ignore any underscored attributes. It no longer recognises Ecto structs, so now all the Ecto stuff is encapsulated in `Kino.Ecto`.


josevalim:
> It still supports a list of structs and we ignore any underscored attributes. It no longer recognises Ecto structs, so now all the Ecto stuff is encapsulated in Kino.Ecto.

Wait, we still handle structs, so we handle Ecto structs. What would be missing?

lib/kino/data_table.ex
```diff
@@ -27,143 +26,91 @@ defmodule Kino.DataTable do
...
-      is_map(record) -> :map
-      key_value_list?(record) -> :key_value_list
-      true -> :other
-    end
+  defp normalize_tabular([%struct{} | _] = tabular) do
+    Enum.map(tabular, fn
+      %^struct{} = item ->
+        item
+        |> Map.from_struct()
+        |> Map.reject(fn {key, _val} ->
```
josevalim, line 68:
```suggestion
        Map.reject(item, fn {key, _val} ->
```

jonatanklosko:
@josevalim yeah, I just meant that we don't use the schema information, they would still be rendered as any other struct :)

josevalim (approved):
Excellent!

Do we want to keep Kino.Ecto, because of the limit+offset navigation? Or would you rather remove it/move it elsewhere?

jonatanklosko:
I think it's still useful because of the pagination, sorting and type information. I think it would make sense to move it elsewere, but we can discuss that once we start splitting into multiple packages.

jonatanklosko:
For the record `Kino.Ecto` uses `@behaviour Kino.Table`, so unless we have all table things in a single package, we may need to make the behaviour public.

PR #124, opened 2022-04-10
Adds layers for chart builder attrs

cristineguadelupe:
This PR strictly solves the persistence issue, assuming that we will always have a single layer plot for a while.

### DIFF
--- a/lib/kino/smart_cell/chart_builder.ex
+++ b/lib/kino/smart_cell/chart_builder.ex
@@ -17,21 +17,23 @@ defmodule Kino.SmartCell.ChartBuilder do
 
   @impl true
   def init(attrs, ctx) do
+    layer = if attrs["layers"], do: List.first(attrs["layers"]), else: nil
+
     fields = %{
-      "chart_type" => attrs["chart_type"] || "bar",
-      "data_variable" => attrs["data_variable"],
+      "chart_title" => attrs["chart_title"],
       "width" => attrs["width"],
       "height" => attrs["height"],
-      "x_field" => attrs["x_field"],
-      "y_field" => attrs["y_field"],
-      "color_field" => attrs["color_field"],
-      "x_field_type" => attrs["x_field_type"],
-      "y_field_type" => attrs["y_field_type"],
-      "color_field_type" => attrs["color_field_type"],
-      "x_field_aggregate" => attrs["x_field_aggregate"],
-      "y_field_aggregate" => attrs["y_field_aggregate"],
-      "color_field_aggregate" => attrs["color_field_aggregate"],
-      "chart_title" => attrs["chart_title"]
+      "chart_type" => layer["chart_type"] || "bar",
+      "data_variable" => layer["data_variable"],
+      "x_field" => layer["x_field"],
+      "y_field" => layer["y_field"],
+      "color_field" => layer["color_field"],
+      "x_field_type" => layer["x_field_type"],
+      "y_field_type" => layer["y_field_type"],
+      "color_field_type" => layer["color_field_type"],
+      "x_field_aggregate" => layer["x_field_aggregate"],
+      "y_field_aggregate" => layer["y_field_aggregate"],
+      "color_field_aggregate" => layer["color_field_aggregate"]
     }
 
     ctx =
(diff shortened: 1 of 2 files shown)


test/kino/smart_cell/chart_builder_test.exs
```diff
@@ -68,13 +78,22 @@ defmodule Kino.SmartCell.ChartBuilderTest do
...
+        update_in(
+          @single_layer["layers"],
+          [Access.at(0)],
+          &Map.merge(&1, %{
+            "chart_type" => "point",
+            "x_field_type" => "nominal",
+            "y_field_type" => "quantitative",
+            "color_field" => "c"
+          })
+        )
```
jonatanklosko, line 91: You can extract merging into a separate function and build attrs like this:

```elixir
def build_attrs(root_attrs, layer_attrs) do
  root_attrs = Map.merge(%{...}, root_attrs)
  layer_attrs = Map.merge(%{...}, layer_attrs)
  Map.put(root_attrs, "layers", [layer_attrs])
end
```

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -3,6 +3,7 @@ defmodule Kino.SmartCell.ChartBuilder do
   use Kino.JS.Live
   use Kino.SmartCell, name: "Chart builder"
 
+  @shared_fields ["chart_title", "width", "height"]
```
jonatanklosko, line 6: Perhaps `root_fields`?

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -102,7 +111,15 @@ defmodule Kino.SmartCell.ChartBuilder do
...
+
+    ctx =
+      if field in @shared_fields do
+        update(ctx, :fields, &Map.put(&1, field, parsed_value))
+      else
+        updated_layers = put_in(ctx.assigns.fields["layers"], [Access.at(0), field], parsed_value)
+        update(ctx, :fields, &Map.put(&1, "layers", updated_layers))
+      end
+
     broadcast_event(ctx, "update", %{"fields" => %{field => parsed_value}})
```
jonatanklosko, line 123: I think this now sends layer field updates that look like top-level field updates on the client.

jonatanklosko:
I dropped some minor comments, but actually it would be easier to only change `init` and `to_attrs` to extract/add the layer, while keeping the state and client unchanged, wdyt?

cristineguadelupe:
> I dropped some minor comments, but actually it would be easier to only change `init` and `to_attrs` to extract/add the layer, while keeping the state and client unchanged, wdyt?

Sounds good!

lib/kino/smart_cell/chart_builder.ex
```diff
@@ -260,4 +264,22 @@ defmodule Kino.SmartCell.ChartBuilder do
...
+  def add_layer(attrs) do
+    layer = Map.drop(attrs, ["chart_title", "width", "height"])
+
+    root = %{
+      "chart_title" => attrs["chart_title"],
+      "width" => attrs["width"],
+      "height" => attrs["height"]
+    }
+
+    Map.merge(root, %{"layers" => [layer]})
```
jonatanklosko, line 277:
```suggestion
    {root, layer} = Map.split(attrs, ["chart_title", "width", "height"])
    Map.put(root, "layers", [layer])
```

test/kino/smart_cell/chart_builder_test.exs
```diff
@@ -43,7 +43,9 @@ defmodule Kino.SmartCell.ChartBuilderTest do
     end
 
     test "source for a basic line plot with alias" do
-      attrs = Map.merge(@defaults, %{"chart_type" => "line", "vl_alias" => Vl})
+      attrs =
+        Map.merge(@defaults, %{"chart_type" => "line", "vl_alias" => Vl})
+        |> ChartBuilder.add_layer()
```
jonatanklosko, line 48: Instead of exposing the function we can add a helper like

```elixir
defp build_attrs(root_attrs, layer_attrs) do
  root_attrs = Map.merge(%{...}, root_attrs)
  layer_attrs = Map.merge(%{...}, layer_attrs)
  Map.put(root_attrs, "layers", [layer_attrs])
end
```

where `merge` first argument are the corresponding defaults :)

jonatanklosko (approved):
Perfect! :cat:

PR #127, opened 2022-04-13
New design

cristineguadelupe:
CleanShot 2022-04-13 at 14 05 05@2x
CleanShot 2022-04-13 at 14 06 09@2x
CleanShot 2022-04-13 at 14 06 20@2x
CleanShot 2022-04-13 at 14 07 20@2x

### DIFF
--- a/lib/assets/chart_builder/main.css
+++ b/lib/assets/chart_builder/main.css
@@ -6,61 +6,31 @@
   --gray-50: #f8fafc;
   --gray-100: #f0f5f9;
   --gray-200: #e1e8f0;
+  --gray-300: #cad5e0;
   --gray-400: #91a4b7;
   --gray-500: #61758a;
   --gray-600: #445668;
   --gray-800: #1c2a3a;
+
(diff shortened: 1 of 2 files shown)


cristineguadelupe:
Maybe we should put the info-box outside the borders...

josevalim:
Yes, the info box should definitely be outside.

lib/assets/chart_builder/main.css
```diff
@@ -6,61 +6,33 @@
...
-  display: flex;
-  flex-direction: row;
-  align-items: flex-start;
-  align-content: stretch;
-  flex-wrap: nowrap;
-}
 
-.row > *:not(:first-child) {
-  margin-left: 8px;
+  --primary-100: #ecf0ff;
```
jonatanklosko, line 15: Just so it matches Livebook :)

```suggestion
  --blue-100: #ecf0ff;
```

jonatanklosko (approved):
Fantastic, just one comment and ship it!

PR #129, opened 2022-04-13
Fix: info box padding

cristineguadelupe:


josevalim (approved):

lib/assets/chart_builder/main.css
```diff
@@ -124,6 +124,10 @@ input[type=number] {
   margin-bottom: 0.3em;
 }
 
+.info-box p:first-child {
+  margin-top: 0;
```
jonatanklosko, line 128: Ah this kills default browser styles, maybe we should have `margin: 0` on the paragraphs and start from there to make sure we don't rely on the defaults?

cristineguadelupe: Done!

jonatanklosko (approved):

PR #135, opened 2022-04-16
Smart Cells - responsiveness

cristineguadelupe:
-- WIP --

- [x] Chart Build
- [x] DB Connection
- [x] SQL

fullsize2

small

extra-small

### DIFF
--- a/lib/assets/db_connection/main.css
+++ b/lib/assets/db_connection/main.css
@@ -79,6 +79,10 @@ input[type=number] {
   min-width: 150px;
 }
 
+.input--xs:is(input[type="text"]) {
+  max-width: 50%;
+}
+
 .input-label {
   display: block;
   margin-bottom: 2px;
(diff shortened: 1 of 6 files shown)


cristineguadelupe:
The minimum width I'm handle is the same as the code editor

CleanShot 2022-04-17 at 00 25 19@2x


https://user-images.githubusercontent.com/5660941/163699204-eec2dc3c-00f8-4bef-8979-0b4fc953e10b.mp4

lib/assets/db_connection/main.css
```diff
@@ -79,6 +79,10 @@ input[type=number] {
   min-width: 150px;
 }
 
+.input--xs:is(input[type="text"]) {
```
jonatanklosko, line 82: I think we should have a separate class that matches the expected behaviour, changing a generic class for specific elements can be quite confusing :)

Sidenote: [:is support](https://caniuse.com/?search=is) is fairly recent.

jonatanklosko (approved):
:rocket:

PR #140, opened 2022-04-19
Add SmartCell to connect with Google BigQuery

aleDsz:
## Demo

image

Query with pure code block:

image

### DIFF
--- a/lib/assets/big_query/main.js
+++ b/lib/assets/big_query/main.js
@@ -1,0 +1,91 @@
+export function init(ctx, info) {
+  ctx.importCSS("main.css");
+  ctx.importCSS("https://fonts.googleapis.com/css2?family=Inter:wght@400;500&display=swap");
+
+  ctx.root.innerHTML = `
+    <div class="app">
+      <div id="info-box" class="info-box"></div>
+      <div class="container">
+        <div class="row header">
+          <div class="field">
+            <label class="input-label"> Assign to </label>
+            <input class="input input--xs input--text" name="variable" type="text" />
+          </div>
+        </div>
+        <div class="row mixed-row">
+          <div class="field grow">
+            <label class="input-label">Project ID</label>
+            <input class="input" name="project_id" type="text" />
+          </div>
+          <div class="field grow">
+            <label class="input-label">Dataset</label>
+            <input class="input" name="dataset" type="text" />
+          </div>
+        </div>
+        <div class="row">
+          <div class="field grow">
+            <label class="input-label">Private Key ID</label>
+            <input class="input" name="private_key_id" type="password" />
+          </div>
+          <div class="field grow">
+            <label class="input-label">Private Key</label>
+            <input class="input" name="private_key" type="password" />
+          </div>
+        </div>
+        <div class="row">
+          <div class="field grow">
+            <label class="input-label">Client E-mail</label>
+            <input class="input" name="client_email" type="text" />
+          </div>
+          <div class="field grow">
+            <label class="input-label">Client ID</label>
+            <input class="input" name="client_id" type="text" />
+          </div>
+        </div>
+        <div class="row">
+          <div class="field grow">
+            <label class="input-label">Client x509 Certificate URL</label>
+            <input class="input" name="client_x509_cert_url" type="text" />
+          </div>
+        </div>
+      </div>
+    </div>
+  `;
+
+  updateInfoBox(info.missing_dep);
+  setValues(info.fields);
+
+  ctx.root.addEventListener("blur", handleFieldChange, true);
+  ctx.root.addEventListener("change", handleFieldChange);
+
+  function handleFieldChange(event) {
+    const { name, value } = event.target;
+
+    ctx.pushEvent("update_field", { field: name, value });
+  }
+
+  ctx.handleEvent("update", ({ fields }) => {
+    setValues(fields);
+  });
+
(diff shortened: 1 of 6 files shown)


lib/assets/big_query/main.js
```diff
@@ -0,0 +1,94 @@
+export function init(ctx, info) {
+  ctx.importCSS("main.css");
```
aleDsz, line 2: Can I import the `main.css` from DbConnection? Since it's a duplicated file from other SmartCell

jonatanklosko: Assets for each `Kino.JS` must be in a separate directory, however we could have `shared.css` and create symlink to it in each directory.

We plan to split kino into several packages before the release, so let's keep it simple for now, we can revisit CSS to reduce duplication later. Also, if we manage to merge this cell with Db connection, we won't have the problem altogether :)

lib/kino/smart_cell/big_query.ex
```diff
@@ -0,0 +1,112 @@
...
+  @impl true
+  def to_source(attrs) do
+    attrs
+    |> to_quoted()
+    |> Kino.Utils.Code.quoted_to_string()
+  end
+
+  defp to_quoted(attrs) do
+    quote do
+      alias unquote(Connection), as: BigQuery
```
jonatanklosko, line 80: An important aspect of smart cells is teaching people how to write the code that they would need to write otherwise, so we should minimize hiding it behind `Kino.*` functions.

Looking at the code, directly we could generate this:

```elixir
scopes = ["https://www.googleapis.com/auth/cloud-platform"]

credentials = %{
  "type" => "...",
  "project_id" => "...",
  "private_key_id" => "...",
  "private_key" => "...",
  "client_email" => "...",
  "client_id" => "...",
  "auth_uri" => "...",
  "token_uri" => "...",
  "auth_provider_x509_cert_url" => "...",
  "client_x509_cert_url" => "..."
}

{:ok, goth} =
  Kino.start_child({Goth, source: {:service_account, credentials, [scopes: scopes]}})

{:ok, %{token: token}} = Goth.fetch(goth)
conn = GoogleApi.BigQuery.V2.Connection.new(token)
```

The only issue is that `conn` is a Finch client, so we can't really tell that it's the db connection. Also, it doesn't include some of the information necessary for making queries. So to integrate it into the SQL cell, we need a more informative struct, which I think gives two options:

* create a plain struct `%Kino.SmartCell.DbConnection.BigQuery{conn: conn}` and put any other relevant information there
* create a separate package specifically for Big Query that makes the HTTP requests directly

@josevalim any other ideas?

lib/assets/big_query/main.js
aleDsz: Nice

lib/kino/smart_cell/big_query.ex
jonatanklosko: Yeah, that's what I had in mind with the second option. It wouldn't be Livebook/Kino specific in any way, but would provide a more straightforward API, with a proper connection struct. Basically it should be something usable on its own that we only use to generate the code. It's better to avoid this kind of of micro-packages, but in this case it would be useful.

aleDsz: > The only issue is that `conn` is a Finch client

This is a bigger issue, because if I get the `{:ok, conn} = Finch...`, conn will be a PID which doesn't work with Finch's library. They asks for the GenServer's name that we defined with `Kino.start_link/1`. So, even with Finch, I need to export a variable (`conn`) with the module name.

```elixir
scopes = ["https://www.googleapis.com/auth/cloud-platform"]

credentials = %{
  "type" => "...",
  "project_id" => "...",
  "private_key_id" => "...",
  "private_key" => "...",
  "client_email" => "...",
  "client_id" => "...",
  "auth_uri" => "...",
  "token_uri" => "...",
  "auth_provider_x509_cert_url" => "...",
  "client_x509_cert_url" => "..."
}

conn = BigQuery

{:ok, goth} = Goth.start_link(source: {:service_account, credentials, scopes: scopes})
{:ok, _finch_pid} = Kino.start_child({Finch, name: BigQuery})
```

And, to request:

```elixir
headers = []
body = nil

request = Finch.build(:get, "https://www.google.com.br", headers, body, [])
{:ok, _response} = Finch.request(request, conn)
```

A simple example, but in my guess, this is less user-friendly than before.

jonatanklosko: > create a plain struct `%Kino.SmartCell.DbConnection.BigQuery{conn: conn}` and put any other relevant information there

Oh, actually it doesn't need to be a struct, just anything that we can clearly recognize as a big query connection. So it could be a regular map like `conn = %{big_query_conn: conn, ...}`. That we can easily match in the SQL cell and generate the code accordingly. That's slightly indirect, but a pretty reasonable option IMO.

aleDsz: This is what needs to be used to perform the same request as the `google_api_big_query` lib

image

jonatanklosko: One other piece to the puzzle is that we definitely need a struct for the response. With the current APIs we have `GoogleApi.BigQuery.V2.Model.QueryRequest`, which is good, otherwise we will need to come up with something. That's because we will soon have the [table](https://github.com/dashbitco/table) package, which means we can implement a protocol for the response struct, and there we will handle all the type conversion, etc. So we won't need the post-response steps in the query code :D

PR #144, opened 2022-04-27
Integrate table reader into chart builder

jonatanklosko:
Updates chart builder to look for any any `Table.Reader.t()` compatible data. Also, passes `:only` to `Vl.data_from_values`, so that the specification includes only the user-selected columns.


cristineguadelupe (approved):
Pretty nice 😊

Issue #154, opened 2022-05-30
Print supervision trees


josevalim:
If a process represents a supervisor (based on Process.info and Supervisor.which_children), we  can use Kino.Markdown with Mermaid to print the supervision tree. For example:

```mermaid
graph TD
    A(#PID<0.23.3>) --> B(Foo.Bar.Sup)
    B ---> C(Foo.Baz.Sup)
    C --->D(One)
    C --->E(#PID<1.2.3>)
    C --->F(Three)
    B -.->F
    style A stroke-width:3px
```

The root has a different stroke to mark the initial PID. Dotted lines can used for links.

josevalim:
Another idea is `Kino.Process.trace/2`:

```
Kino.Process.trace(pid, fn ->

end)
```

The above will trace all messages received by `pid` (or list of pids) and build a sequence diagram.

akoutmos:
I am getting close to opening a PR for this but had a question regarding links. Should a process with a link to another process in a separate supervision tree be added to the graph? Or should links that are only present between children of the root supervisor be drawn out? (hopefully that question makes sense)

Also, how do you feel about color coding supervisors versus workers?

![image](https://user-images.githubusercontent.com/4753634/175759983-0dfa16a7-13ad-4924-aa41-63397623da64.png)

josevalim:
Exciting!!!!

Only links between processes already in the tree which _are not_ direct children. So in your case, no links between children and their parents, but a link between 0.271.0 and 0.269.0.

josevalim:
Btw, let's call the module `Kino.Process`. :)

PR #158, opened 2022-06-26
Adding supervision tree visualizer

akoutmos:
This PR addresses half of #154. It supports rendering supervision trees top-down and left to right. You can render the supervisor by name or by PID. The links that are displayed are only between processes within the supervision tree and process links external to the supervision tree are omitted.

I haven't added tests yet, but will add them once I know I am on the right track :smile:. Thanks in advance for the review!

Some sample images for reference:

image
image

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -1,0 +1,138 @@
+defmodule Kino.Process do
+  @moduledoc """
+
+  ## Examples
+  """
+
+  defmodule SupervisorGraphNode do
+    @moduledoc false
+
+    defstruct [:id, :pid, :type]
+
+    @doc false
+    def new(id, pid, type) do
+      %__MODULE__{
+        id: id,
+        pid: pid,
+        type: type
+      }
+    end
+
+    @doc false
+    def generate_node(%__MODULE__{id: id, pid: pid, type: type}) do
+      type =
+        if id == 0 do
+          :root
+        else
+          type
+        end
+
+      display =
+        pid
+        |> Process.info(:registered_name)
+        |> case do
+          {:registered_name, []} -> inspect(pid)
+          {:registered_name, name} -> inspect(name)
+        end
+
+      "#{id}(#{display}):::#{type}"
+    end
+  end
+
+  defmodule SupervisorGraphEdge do
+    @moduledoc false
+
+    defstruct [:node_1, :node_2, :relationship]
+
+    @doc false
+    def new(node_1, node_2, relationship) do
+      %__MODULE__{
+        node_1: node_1,
+        node_2: node_2,
+        relationship: relationship
+      }
+    end
+
+    @doc false
+    def generate_mermaid_entry(%__MODULE__{node_1: node_1, node_2: node_2, relationship: :link}) do
+      "#{SupervisorGraphNode.generate_node(node_1)} -..- #{SupervisorGraphNode.generate_node(node_2)}"
+    end
+
+    def generate_mermaid_entry(%__MODULE__{
+          node_1: node_1,
+          node_2: node_2,
+          relationship: :supervisor
+        }) do
+      "#{SupervisorGraphNode.generate_node(node_1)} ---> #{SupervisorGraphNode.generate_node(node_2)}"
+    end
+  end
+
+  @doc """
+  Generates a Mermaid.js graph of the supervision tree
+  """
+  @spec generate_supervision_tree(pid()) :: Kino.Markdown.t()
+  def generate_supervision_tree(supervisor) when is_pid(supervisor) do
+    initial_pid_lookup = Map.put(%{}, supervisor, 0)
+
+    {edges, _, _} =
+      supervisor
+      |> Supervisor.which_children()
+      |> traverse_processes(supervisor, {[], 1, initial_pid_lookup})
+      |> traverse_links(supervisor)
+
+    edges =
+      Enum.map_join(edges, "\n", fn edge -> SupervisorGraphEdge.generate_mermaid_entry(edge) end)
+
+    Kino.Markdown.new("""
+    ```mermaid
+    graph TD;
+    #{edges}
+    classDef root fill:#c4b5fd, stroke:#374151, stroke-width:4px;
+    classDef supervisor fill:#c4b5fd, stroke:#374151, stroke-width:1px;
+    classDef worker fill:#93c5fd, stroke:#374151, stroke-width:1px;
+    ```
+    """)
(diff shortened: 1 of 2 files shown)


lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+        |> case do
+          {:registered_name, []} -> inspect(pid)
+          {:registered_name, name} -> inspect(name)
+        end
+
+      "#{id}(#{display}):::#{type}"
+    end
+  end
+
+  defmodule SupervisorGraphEdge do
```
josevalim, line 76: Given these modules are private, we don't need a module, you can use "private" maps instead:

    defp graph_edge(node_1, node_2, relationship) do
      %{node_1: node_1, node_2: node_2, relationship: relationship}
    end

lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+
+  ## Options
+
+    * `:direction` - defines the direction of the Mermaid.js graph. The
+      value can either be `:td` (top-down) or `:lr` (left to right) with
+      the default being `:td`. Optional
+
+  """
+  @spec generate_supervision_tree(supervisor(), keyword()) :: Markdown.t()
+  def generate_supervision_tree(supervisor, opts \\ [])
```
josevalim, line 118: Can we rename this to `sup_tree`? :) Or at least `supervision_tree` (but I would go with the former... it is fine short names here as it is an exploratory library.

lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+
+  def generate_supervision_tree(named_supervisor, opts) when is_atom(named_supervisor) do
+    case Process.whereis(named_supervisor) do
```
josevalim, line 121: Move the pid clause first and then do something like this:

```suggestion
  def generate_supervision_tree(named_supervisor, opts) do
    case GenServer.whereis(named_supervisor) do
```

This will also allow us to support via tuples. :)

lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+        raise ArgumentError, "the provided PID #{inspect(supervisor)} is not a supervisor"
+    end
+  end
+
+  def generate_supervision_tree(invalid_pid, _opts) do
+    raise ArgumentError, "expected a PID, got: #{inspect(invalid_pid)}"
+  end
+
+  defp convert_direction(:td), do: "TD"
+  defp convert_direction(:lr), do: "LR"
```
josevalim, line 172:
```suggestion
  defp convert_direction(:top_down), do: "TD"
  defp convert_direction(:left_right), do: "LR"
```

?

lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+          node_1: node_1,
+          node_2: node_2,
+          relationship: :supervisor
+        }) do
+      "#{SupervisorGraphNode.generate_node(node_1)} ---> #{SupervisorGraphNode.generate_node(node_2)}"
+    end
+  end
+
+  @doc """
+  Generates a Mermaid.js graph of a supervision tree. The provided supervisor can be either a named
```
josevalim, line 105: First paragraph of the docs should be a short summary. :)

```suggestion
  Generates a Mermaid.js graph of a supervision tree.

  The provided supervisor can be either a named
```

lib/kino/process.ex
```diff
@@ -0,0 +1,251 @@
...
+  process or a PID. The supervision tree is displayed with solid lines denoting supervisor-worker
+  relationships and dotted lines denoting links between processes. The graph rendering supports
+  the following options:
+
+  ## Options
+
+    * `:direction` - defines the direction of the Mermaid.js graph. The
+      value can either be `:td` (top-down) or `:lr` (left to right) with
+      the default being `:td`. Optional
```
jonatanklosko, line 114:
```suggestion
      value can either be `:td` (top-down) or `:lr` (left to right). Defaults
      to `:td`
```

akoutmos:
Thanks for the feedback! I think I addressed all of the comments regarding the new `sup_tree` function. I will work on implementing `app_tree` as well later today :smiley:.

lib/kino/process.ex
```diff
@@ -0,0 +1,229 @@
...
+      graph #{direction};
+      #{edges}
+      classDef root fill:#c4b5fd, stroke:#374151, stroke-width:4px;
+      classDef supervisor fill:#c4b5fd, stroke:#374151, stroke-width:1px;
+      classDef worker fill:#93c5fd, stroke:#374151, stroke-width:1px;
+      ```
+      """)
+    catch
+      _, _ ->
+        raise ArgumentError, "the provided PID #{inspect(supervisor)} is not a supervisor"
```
jonatanklosko, line 100: Can we wrap the `Supervisor.which_children` call specifically? Otherwise it's not obvious where this comes from :)

akoutmos:
Almost done with the `app_tree` function as well. Should the `:application_master` processes get any kind of special formatting or is the image below good? I also left the thicker border on the application supervisor.

image

lib/kino/process.ex
```diff
@@ -0,0 +1,305 @@
...
+                "the provided identifier #{inspect(supervisor)} does not reference a running process"
+      end
+
+    supervisor_children =
+      try do
+        Supervisor.which_children(supervisor)
+      catch
+        _, _ ->
+          raise ArgumentError, "the provided process #{inspect(supervisor)} is not a supervisor"
+      end
```
jonatanklosko, line 172: Sidenote: since this does GenServer call, it can easily crash or block the receiver, so even though we catch an error, there's likely an undesired side effect. @josevalim can we check for a supervisor upfront, perhaps by looking at `$initial_call`, or is that unreliable?

lib/kino/process.ex
```diff
@@ -0,0 +1,305 @@
...
+
+    direction = direction_from_opts(opts)
+    edges = traverse_supervisor(root_supervisor)
+
+    [ancestor] =
+      root_supervisor
+      |> Process.info()
+      |> Keyword.get(:dictionary)
+      |> Keyword.get(:"$ancestors")
```
jonatanklosko, line 68:
```suggestion
    {:dictionary, dictionary} = Process.info(root_supervisor, :dictionary)
    [ancestor] = dictionary[:"$ancestors"]
```

josevalim:

jonatanklosko:
@akoutmos fantastic job! :cat:

akoutmos:
Thanks for helping me get this feature in :)

PR #160, opened 2022-06-27
Make data traversal lazy in Kino.DataTable

jonatanklosko:
Uses `Enumerable.slice` on row data if available, otherwise takes and caches enumerable items on demand.

This relies on some small changes to `Table`, I'm opening this for easier discussion.

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -51,14 +51,18 @@ defmodule Kino.DataTable do
     sorting_enabled = Keyword.get(opts, :sorting_enabled, true)
     keys = opts[:keys]
 
-    {data_rows, data_columns, count} =
+    {_, meta, _} = reader = init_reader!(tabular)
+
+    count = meta[:count] || infer_count(reader, tabular)
+
+    {data_rows, data_columns} =
       if keys do
-        {rows, info} = Table.to_rows_with_info(tabular, only: keys)
-        nonexistent = keys -- info.columns
-        {rows, keys -- nonexistent, info.count}
+        rows = Table.to_rows(reader, only: keys)
+        nonexistent = keys -- meta.columns
+        {rows, keys -- nonexistent}
       else
-        {rows, info} = Table.to_rows_with_info(tabular)
-        {rows, info.columns, info.count}
+        rows = Table.to_rows(reader)
+        {rows, meta.columns}
       end
 
     Kino.Table.new(__MODULE__, {data_rows, data_columns, count, name, sorting_enabled})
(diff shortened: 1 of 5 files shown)


lib/kino/data_table.ex
```diff
@@ -49,18 +49,19 @@ defmodule Kino.DataTable do
...
-        {rows, keys -- nonexistent}
+    {data_rows, data_columns, count} =
+      if keys do
+        {rows, info} = Table.to_rows_with_info(tabular, only: keys)
+        nonexistent = keys -- info.columns
+        {rows, keys -- nonexistent, info.count}
       else
-        {rows, %{columns: columns}} = Table.to_rows_with_info(tabular)
-        {rows, columns}
+        {rows, info} = Table.to_rows_with_info(tabular)
```
jonatanklosko, line 60: Currently `info` is not exactly the metadata from `Table.Reader.init`, it's a [normalized map](https://github.com/dashbitco/table/blob/950537310ce490d0169c2d5969e00bfd0f02f390/lib/table.ex#L54). We can add `:count` there to be either integer or nil. Alternatively we can return the whole metadata, @josevalim wdyt?

lib/kino/data_table.ex
```diff
@@ -79,39 +80,114 @@ defmodule Kino.DataTable do
...
+       total_rows: total_rows
+     }, %{state | total_rows: total_rows, slicing_cache: slicing_cache}}
+  end
 
-    Enum.slice(sorted_data, rows_spec.offset, rows_spec.limit)
+  defp query(data, slicing_fun, slicing_cache, rows_spec) do
+    if order_by = rows_spec[:order_by] do
+      sorted = Enum.sort_by(data, & &1[order_by], rows_spec.order)
+      records = Enum.slice(sorted, rows_spec.offset, rows_spec.limit)
+      {records, Enum.count(sorted), slicing_cache}
```
jonatanklosko, line 188: This path always loads all the data.

lib/kino/data_table.ex
```diff
@@ -49,18 +49,19 @@ defmodule Kino.DataTable do
...
+      if keys do
+        {rows, info} = Table.to_rows_with_info(tabular, only: keys)
+        nonexistent = keys -- info.columns
+        {rows, keys -- nonexistent, info.count}
       else
-        {rows, %{columns: columns}} = Table.to_rows_with_info(tabular)
-        {rows, columns}
+        {rows, info} = Table.to_rows_with_info(tabular)
+        {rows, info.columns, info.count}
       end
```
jonatanklosko, line 62: @josevalim for the built-in list `Table.Reader` implementation, we can't special case by calling `init` and checking the enumerable, because the enumerable is actually a stream that normalizes maps/keyval lists into list of columns.

One way to solve this would be adding `count: length(...)` to the metadata, since the protocol implementation knows that all the data is in memory. The downside is that we need to traverse all rows, which we wanted to avoid initially.

Alternatively we could have a flag that indicates if the data is in memory or lazily loaded, and then we know if we can do `Enum.count` ourselves. Though this doesn't feel elegant to me.

lib/kino/data_table.ex
josevalim: Hrm... my preference would be to return the metadata as is. Why do we have to normalize it?

lib/kino/data_table.ex
josevalim: Right... that's why I think we should call `Table.Reader.init` directly and allow the result of `init` to be given to `to_rows` and friends.

jonatanklosko: Yeah, but my point is that even calling `Table.Reader.init` we don't know if we can call `Enum.count`, because the enumerable is a stream, not a list.

lib/kino/data_table.ex
jonatanklosko: I don't think there's a good reason, other than making the info structure more predictable for the end user.

lib/kino/data_table.ex
```diff
@@ -78,40 +83,153 @@ defmodule Kino.DataTable do
 
   defp normalize_tabular(tabular), do: tabular
 
+  defp init_reader!(tabular) do
+    with :none <- Table.Reader.init(tabular) do
+      raise ArgumentError, "expected valid tabular data, but got: #{inspect(tabular)}"
+    end
+  end
```
jonatanklosko, line 90: Should we expose `Table.init_reader!(tabular)`?

josevalim: I would say no, because we most often can raise better error messages by doing it ourselves.

lib/kino/data_table.ex
```diff
@@ -78,40 +83,153 @@ defmodule Kino.DataTable do
...
+
+  defp infer_count({_, %{count: count}, _}, _), do: count
+
+  defp infer_count({:rows, _, _}, tabular) when is_list(tabular), do: length(tabular)
+
+  defp infer_count({:columns, _, _}, [{_, series} | _]) when is_list(series), do: length(series)
+
+  defp infer_count({:columns, _, _}, %{} = tabular) do
+    {_, series} = Enum.at(tabular, 0)
+    length(series)
```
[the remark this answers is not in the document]
jonatanklosko: It's the same as rows, we can remove this on Elixir v1.14, because then `Enumerable.count(series)` (where series is a list) will give us the count.

PR #162, opened 2022-06-29
Improve edge cases in Kino.DataTable

jonatanklosko:
:cat2:

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -101,8 +101,10 @@ defmodule Kino.DataTable do
   defp infer_count({:columns, _, _}, [{_, series} | _]) when is_list(series), do: length(series)
 
   defp infer_count({:columns, _, _}, %{} = tabular) do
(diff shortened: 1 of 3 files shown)


josevalim:
We need some tests and we are good to go!

lib/kino/data_table.ex
```diff
@@ -113,12 +115,10 @@ defmodule Kino.DataTable do
   end
 
   defp infer_count({:columns, _, enum}, _) do
-    series = Enum.at(enum, 0)
-
-    case Enumerable.count(series) do
-      {:ok, count} -> count
-      _ -> nil
-    end
+    with {:ok, series} <- Enum.fetch(enum, 0),
```
jonatanklosko, line 118: In practice this actually should never be the case, for `:rows` if the data is empty we don't know the columns, but for `:columns` if the data is empty we generally know the column names, it's just that they have no values. Doesn't hurt to use the safe access though.

Issue #159, opened 2022-06-27
Add `Kino.Process.trace/2`


josevalim:
Another idea is `Kino.Process.trace/2`:

```
Kino.Process.trace(pid, fn ->

end)
```

The above will trace all messages received by `pid` (or list of pids) and build a sequence diagram.

josevalim:
Implementation wise, we should either use :erlang.trace or [seq_trace](https://www.erlang.org/doc/man/seq_trace.html). We need to do some proof of concepts to see what works. A very neat example could be Task.async_stream and see how it spawns a process and sends messages and processes terminates and stuff. :)

@akoutmos, let us know if you want to give it a try!

akoutmos:
I put together a really simple PoC using `seq_trace` and ran the following code during a trace:

```elixir
1..5
|> Task.async_stream(fn num ->
  100 * num
end,
max_concurrency: 5)
|> Stream.run()
```

The connections in the sequence diagrams are formatted  using the following code as not to overflow the Mermaid diagram:

```elixir
formatted_message = 
    case message do
      {:EXIT, _, reason} -> "EXIT - normal"
      {:spawn_request, _, _, _, _, _, _, _} -> "spawn_request"
      {:spawn_reply, _, status, _} -> "spawn_reply - #{status}"
      {:down, _, status} -> "down - #{status}"
      {:spawn, _} -> "spawn"
      _ -> ""
    end
```

Is this sort of what you had in mind for this feature?

![image](https://user-images.githubusercontent.com/4753634/176791525-55f2d851-9a5d-4bdd-b116-40b9983494b8.png)

josevalim:
Yes, this is pretty much the building block! It is awesome! Some questions:

1. can we make the timeline start as soon as the line receives the spawn event, or do they always start at the top?
2. similarly, can we make them stop on exit?
3. You can definitely hide stream_reply as that is internal, everything else you can inspect but probably trim the characters print?
4. Do you know what is the last process at the end?
5. Also let’s show the registered name if available :)
6. Can you please PR it as Process.seq_trace? :)

akoutmos:
I'll clean up what I have and open up a PR tomorrow. Below are the answers to your questions:

[...]

PR #165, opened 2022-07-02
Adding process tracing

akoutmos:
This PR implements #159 

Continuing the discussion from that issue:

> Should we not show the bars if the process already exists or should we show the bar from beginning to end?

At the moment I have the bar starting when a process is spawned and stopped when it exits (both of which have to be present during the trace in order for the bar to appear). In other words, it doesn't look like Mermaid renders anything unless there is both an actor activate and deactivate. That and it doesn't look like we can start bars at arbitrary points on the sequence. Rather it looks like they have to start and stop on events. All this to say, *I think* if we want to keep the bars, they'll only be present when a process starts and stops during the trace.

> Regarding the messages, can we do any sort of mouse over thing? Keep only the label and when you mouse over, you see the whole message?

I'll have to explore this a bit more. I tried setting up a pop up [Actor Menu](https://mermaid-js.github.io/mermaid/#/sequenceDiagram?id=actor-menus) as described in the docs...but the menu never popped up lol.

> There are some dotted arrows after the messages... is this the "receive" event? If so, I think we can ignore it. Another option would be for the arrows to not be horizontal, but point to when the messages are received in the other process timeline, but that may be too complicated?

I think the Mermaid diagrams only allow for horizontal lines and cannot be linked back to another event. Do you want me to remove all receive events and just have send events? Or only in certain circumstances? Here is an example with no receives for reference:

image

>We also need to check if we trace :DOWN messages from Process.monitor.

It has been done! I also added that to the example code :).

I have a couple more things that I need to clean up and fix so I am opening this up as a draft PR. Should have those wrapped up during the weekend :). Thanks in advance for the review and have a great weekend!

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -250,8 +250,7 @@ defmodule Kino.Process do
 
     # Get all the participating actors in the trace along with their sequence diagram IDs
     {participants_lookup, _idx} =
-      trace_events
-      |> Enum.reduce({%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
+      Enum.reduce(trace_events, {%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
         acc
         |> maybe_add_participant(from)
         |> maybe_add_participant(to)
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -157,9 +157,9 @@ defmodule Kino.Process do
         Process.monitor(agent_pid)
 
         1..2
-        |> Task.async_stream(fn thing ->
-          Agent.get(agent_pid, fn thing -> thing end)
-          100 * thing
+        |> Task.async_stream(fn value ->
+          Agent.get(agent_pid, fn value -> value end)
+          100 * value
         end,
         max_concurrency: 3)
         |> Stream.run()
@@ -176,9 +176,9 @@ defmodule Kino.Process do
 
       Kino.Process.seq_trace(agent_pid, fn ->
         1..2
-        |> Task.async_stream(fn thing ->
-          Agent.get(agent_pid, fn thing -> thing end)
-          100 * thing
+        |> Task.async_stream(fn value ->
+          Agent.get(agent_pid, fn value -> value end)
+          100 * value
         end,
         max_concurrency: 3)
         |> Stream.run()
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -202,28 +202,26 @@ defmodule Kino.Process do
     previous_tracer = :seq_trace.set_system_tracer(tracer_pid)
 
     # Run the user supplied function and capture the events if no errors were encountered
-    try do
-      # Run the user provided function
-      trace_function.()
-    rescue
-      error ->
+    raw_trace_events =
+      try do
+        try do
+          # Run the user provided function
+          trace_function.()
+        after
+          # Reset all of the seq_trace options
+          :seq_trace.set_system_tracer(previous_tracer)
+          :seq_trace.reset_trace()
+        end
+
+        Tracer.get_trace_events(tracer_pid)
+      after
         # The Tracer GenServer is no longer needed, shut it down
         GenServer.stop(tracer_pid)
-
-        # Reset all of the tracing options
-        :seq_trace.set_system_tracer(previous_tracer)
-        :seq_trace.reset_trace()
-        reraise error, __STACKTRACE__
-    end
-
-    # Reset all of the tracing options
-    :seq_trace.set_system_tracer(previous_tracer)
-    :seq_trace.reset_trace()
+      end
 
     # Get all of the events from the Tracer GenServer
     trace_events =
-      tracer_pid
-      |> Tracer.get_trace_events()
+      raw_trace_events
       |> Enum.filter(fn
         # Skip :spawn_reply messages
         {_, _, _, _, {:spawn_reply, _, _, _}} ->
@@ -244,9 +242,6 @@ defmodule Kino.Process do
       |> Enum.sort_by(fn {_type, timestamp, _from, _to, _message} ->
         timestamp
       end)
-
-    # The Tracer GenServer is no longer needed, shut it down
-    GenServer.stop(tracer_pid)
 
     # Get all the participating actors in the trace along with their sequence diagram IDs
     {participants_lookup, _idx} =
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -200,7 +200,7 @@ defmodule Kino.Process do
   def seq_trace(trace_pids, trace_function) when is_list(trace_pids) or trace_pids == :all do
     # Set up the process message tracer and the Erlang seq_trace_module
     calling_pid = self()
-    {:ok, tracer_pid} = Tracer.start_link(nil)
+    {:ok, tracer_pid} = Tracer.start_link()
     :seq_trace.set_token(:send, true)
     :seq_trace.set_token(:receive, true)
     :seq_trace.set_token(:monotonic_timestamp, true)
@@ -229,28 +229,28 @@ defmodule Kino.Process do
       raw_trace_events
       |> Enum.filter(fn
         # Skip :spawn_reply messages
-        {_, _, _, _, {:spawn_reply, _, _, _}} ->
+        %{message: {:spawn_reply, _, _, _}} ->
           false
 
         # Skip loopback messages
-        {_, _, pid, pid, _message} ->
+        %{from: pid, to: pid} ->
           false
 
         # Filter out messages based on the trace target
-        {_, _, from_pid, to_pid, _message} ->
+        %{from: from_pid, to: to_pid} ->
           trace_pids == :all or from_pid in trace_pids or to_pid in trace_pids
 
         # Reject the rest
         _ ->
           false
       end)
-      |> Enum.sort_by(fn {_type, timestamp, _from, _to, _message} ->
+      |> Enum.sort_by(fn %{timestamp: timestamp} ->
         timestamp
       end)
 
     # Get all the participating actors in the trace along with their sequence diagram IDs
     {participants_lookup, _idx} =
-      Enum.reduce(trace_events, {%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
+      Enum.reduce(trace_events, {%{}, 0}, fn %{from: from, to: to}, acc ->
         acc
         |> maybe_add_participant(from)
         |> maybe_add_participant(to)
@@ -269,7 +269,7 @@ defmodule Kino.Process do
     # Generate the mermaid formatted list of message events
     {formatted_messages, _} =
       trace_events
-      |> Enum.reduce({[], MapSet.new()}, fn {_type, _timestamp, from, to, message},
+      |> Enum.reduce({[], MapSet.new()}, fn %{from: from, to: to, message: message},
                                             {events, started_processes} ->
         events = [normalize_message(from, to, message, participants_lookup) | events]
 
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -188,6 +188,7 @@ defmodule Kino.Process do
   """
   @spec seq_trace(trace_target(), (() -> any())) :: Markdown.t()
   def seq_trace(trace_target \\ :all, trace_function)
+
   def seq_trace(pid, trace_function) when is_pid(pid) do
     seq_trace([pid], trace_function)
   end
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -5,8 +5,10 @@ defmodule Kino.Process do
   """
 
   alias Kino.Markdown
+  alias Kino.Process.Tracer
 
   @type supervisor :: pid() | atom()
+  @type trace_target :: :all | pid() | [pid()]
 
   @doc """
   Generates a visualization of an application tree.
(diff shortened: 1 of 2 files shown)


lib/kino/process.ex
```diff
@@ -137,6 +139,205 @@ defmodule Kino.Process do
...
+    :seq_trace.set_token(:send, true)
+    :seq_trace.set_token(:receive, true)
+    :seq_trace.set_token(:monotonic_timestamp, true)
+    previous_tracer = :seq_trace.set_system_tracer(tracer_pid)
+
+    # Run the user supplied function
+    trace_function.()
+
+    # Reset all of the tracing options
+    :erlang.trace(:all, false, [:send, :receive])
```
josevalim, line 206: I think we don't need to change Erlang trace, do we?

lib/kino/process.ex
```diff
@@ -137,6 +139,205 @@ defmodule Kino.Process do
...
+    :seq_trace.set_system_tracer(previous_tracer)
+    :seq_trace.reset_trace()
```
josevalim, line 208: We should likely wrap the function in a `try` and execute this in `after`. :)

lib/kino/process.ex
```diff
@@ -137,6 +139,205 @@ defmodule Kino.Process do
...
+    end
+  end
+
+  defp normalize_message(_, _, _, _, _), do: ""
+
+  defp label_from_message(message) do
+    case message do
+      {:EXIT, _, reason} -> "EXIT: #{label_from_reason(reason)}"
+      {:spawn_request, _, _, _, _, _, _, _} -> "SPAWN"
+      {:down, _, reason} -> "DOWN: #{label_from_reason(reason)}"
```
josevalim, line 321:
```suggestion
      {:DOWN, _, :process, _, reason} -> "DOWN: #{label_from_reason(reason)}"
```

The label that you were before was specific to Task.async_stream. :)

lib/kino/process.ex
```diff
@@ -137,6 +139,205 @@ defmodule Kino.Process do
...
+  defp label_from_reason(:normal), do: "normal"
+  defp label_from_reason(:shutdown), do: "shutdown"
+  defp label_from_reason({:shutdown, _}), do: "shutdown"
+  defp label_from_reason(_), do: "abnormal"
+
+  defp label_from_value(tuple)
+       when is_tuple(tuple) and is_atom(elem(tuple, 0)),
+       do: elem(tuple, 0)
+
+  defp label_from_value(atom) when is_atom(atom), do: atom
```
josevalim, line 337: One more:

```suggestion
  defp label_from_value(ref) when is_reference(ref), do: inspect(ref)
```

lib/kino/process.ex
josevalim: Likely something like this:

```elixir
{res, trace_events} =
  try do
    {trace_function.(), Tracer.get_trace_events(tracer_pid)}
  after
    Tracer.stop(tracer_pid)
    :seq_trace.set_system_tracer(previous_tracer)
    :seq_trace.reset_trace()
  end
```

So you ensure no seq_trace nor process leak.

josevalim:
@jonatanklosko the current implementation discards the result of the function given to trace and I don't like that. We could instead return `{res, kino_markdown}`, but then the result of course won't be printed. So here is my suggestion:

1. Have this function return both `{res, kino_markdown}`
2. Also add `Kino.Process.render_trace`, that calls `trace/2`, renders the trace behind the scenes, and returns only `res`

WDYT?

josevalim (commented):
Amazing @akoutmos! I have added only some minor final comments!

> Do you want me to remove all receive events and just have send events? 

Yes, I think we can remove them all. :)

lib/kino/process.ex
akoutmos: Ooops. That was copy and paste carry over from when I was experimenting with trace versus seq_trace. Fixed!

lib/kino/process.ex
```diff
@@ -137,6 +139,212 @@ defmodule Kino.Process do
...
+
+        Agent.stop(agent_pid)
+      end)
+  """
+  @spec seq_trace(trace_target(), (() -> any())) :: Markdown.t()
+  def seq_trace(pid, trace_function) when is_pid(pid) do
+    seq_trace([pid], trace_function)
+  end
+
+  def seq_trace(trace_pids, trace_function) do
```
josevalim, line 194:
```suggestion
  def seq_trace(trace_pids, trace_function) when is_list(trace_pids) do
```

lib/kino/process.ex
```diff
@@ -137,6 +139,212 @@ defmodule Kino.Process do
...
+
+          {:registered_name, name} ->
+            "participant #{idx} AS #{inspect(name)};"
+        end
+      end)
+
+    # Generate the mermaid formatted list of message events
+    messages =
+      trace_events
+      |> Enum.map_join("\n", fn {type, _timestamp, from, to, message} ->
```
josevalim, line 272: It is a style thing but can you please remove the "single pipe" pipelines? More than one `|>` is fine tho. :)

lib/kino/process.ex
```diff
@@ -137,6 +139,212 @@ defmodule Kino.Process do
...
+        case Process.info(pid, :registered_name) do
+          nil ->
+            "participant #{idx} AS #35;PID#{:erlang.pid_to_list(pid)};"
+
+          {:registered_name, []} ->
+            "participant #{idx} AS #35;PID#{:erlang.pid_to_list(pid)};"
+
+          {:registered_name, name} ->
+            "participant #{idx} AS #{inspect(name)};"
+        end
```
josevalim, line 266:
```suggestion
        case Process.info(pid, :registered_name) do
          {:registered_name, name} when is_atom(name) ->
            "participant #{idx} AS #{inspect(name)};"

          _ ->
            "participant #{idx} AS #35;PID#{:erlang.pid_to_list(pid)};"
        end
```

lib/kino/process.ex
```diff
@@ -137,6 +139,212 @@ defmodule Kino.Process do
     """)
   end
 
+  @doc """
+  Generate a sequence diagram of process messages.
```
josevalim, line 143: Now that you added `:all` and pids, it is important to clarify that the trace starts from the the current process.

```suggestion
  Generate a sequence diagram of process messages starting from `self()`.
```

lib/kino/process.ex
```diff
@@ -137,6 +139,212 @@ defmodule Kino.Process do
...
+          Agent.get(agent_pid, fn thing -> thing end)
+          100 * thing
+        end,
+        max_concurrency: 3)
+        |> Stream.run()
+
+        Agent.stop(agent_pid)
+      end)
+  """
+  @spec seq_trace(trace_target(), (() -> any())) :: Markdown.t()
```
josevalim, line 189:
```suggestion
  @spec seq_trace(trace_target(), (() -> any())) :: Markdown.t()
  def seq_trace(trace_target \\ :all, trace_function)
```

?

akoutmos: The lazy programmer in me likes that :D

lib/kino/process.ex
```diff
@@ -137,6 +139,210 @@ defmodule Kino.Process do
...
+      |> Enum.reduce({%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
+        acc
+        |> maybe_add_participant(from)
+        |> maybe_add_participant(to)
+      end)
+
+    # Generate the Mermaid formatted list of participants
+    participants =
+      participants_lookup
+      |> Enum.map_join("\n", fn {pid, idx} ->
```
josevalim, line 258:
```suggestion
      Enum.map_join(participants_lookup, "\n", fn {pid, idx} ->
```

lib/kino/process.ex
```diff
@@ -137,6 +139,210 @@ defmodule Kino.Process do
...
+          false
+      end)
+      |> Enum.sort_by(fn {_type, timestamp, _from, _to, _message} ->
+        timestamp
+      end)
+
+    # Get all the participating actors in the trace along with their sequence diagram IDs
+    {participants_lookup, _idx} =
+      trace_events
+      |> Enum.reduce({%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
```
josevalim, line 249:
```suggestion
      Enum.reduce(trace_events, {%{}, 0}, fn {_type, _timestamp, from, to, _message}, acc ->
```

josevalim (approved):

akoutmos:
Thanks for the review @josevalim! Really appreciate it. I reworked the exception handling a little so that the call to the tracer GenServer was not part of the trace capture. I also need to rework that last reduce to handle the scenario when a process is started prior to the trace, but terminated during the trace. Mermaid throws an error if you attempt to deactivate an actor without it being activated prior.

After that I think it is good to go :). Thanks!

josevalim:
You can also do nested trys. The nesting is uglier but less code duplication and perhaps easier to understand?

```elixir
try do
  try do
    fun.()
  after
    stop_trace
  end

  get_trace_events
after
  GenServer.stop
end
```

akoutmos:
> You can also do nested trys. The nesting is uglier but less code duplication and perhaps easier to understand?
> 
> ```elixir
> try do
>   try do
>     fun.()
>   after
>     stop_trace
>   end
> 
>   get_trace_events
> after
>   GenServer.stop
> end
> ```

It has been done :)

akoutmos:
I fixed that last issue that was on my TODO list and switched the PR from draft to ready. Let me know if there are any other issues I should address.

I think the only open question at the moment is whether the result is kept from the trace function?

josevalim:
I will let you know soon. I also just realized that we could name the current process as `self()` in the diagram. WDYT?

akoutmos:
> I will let you know soon. I also just realized that we could name the current process as `self()` in the diagram. WDYT?

Sounds good :). I updated how the participating PIDs are generated and the calling PID now show up as `self()`. I think that makes it nice and clear for users following the trace.

josevalim:
Hi @akoutmos, let's please change this one to return `{res, kino}`. Then we will add render_app_tree, render_sup_tree and render_seq_trace that render as a side-effect. The first two will return `Kino.nothing`, the last one will return the result of the function. WDYT? Can you please push those final changes? :)

lib/kino/process.ex
```diff
@@ -137,6 +139,229 @@ defmodule Kino.Process do
...
+      {:ok, agent_pid} = Agent.start_link(fn -> [] end)
+      Process.monitor(agent_pid)
+
+      Kino.Process.seq_trace(agent_pid, fn ->
+        1..2
+        |> Task.async_stream(fn value ->
+          Agent.get(agent_pid, fn value -> value end)
+          100 * value
+        end,
+        max_concurrency: 3)
```
jonatanklosko, line 183:
```suggestion
        |> Task.async_stream(
          fn value ->
            Agent.get(agent_pid, fn value -> value end)
            100 * value
          end,
          max_concurrency: 3
        )
```

lib/kino/process.ex
```diff
@@ -137,6 +139,229 @@ defmodule Kino.Process do
...
+      Kino.Process.seq_trace(fn ->
+        {:ok, agent_pid} = Agent.start_link(fn -> [] end)
+        Process.monitor(agent_pid)
+
+        1..2
+        |> Task.async_stream(fn value ->
+          Agent.get(agent_pid, fn value -> value end)
+          100 * value
+        end,
+        max_concurrency: 3)
```
jonatanklosko, line 164:
```suggestion
        |> Task.async_stream(
          fn value ->
            Agent.get(agent_pid, fn value -> value end)
            100 * value
          end,
          max_concurrency: 3
        )
```

lib/kino/process/tracer.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Kino.Process.Tracer do
+  @moduledoc false
+
+  use GenServer
+
+  def start_link(_) do
```
jonatanklosko, line 6: Since we start it by hand, we don't need an artificial argument :)

```suggestion
  def start_link() do
```

lib/kino/process/tracer.ex
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  @impl true
+  def handle_call(:get_trace_events, _from, trace_events) do
+    {:reply, trace_events, trace_events}
+  end
+
+  @impl true
+  def handle_info({:seq_trace, _, {:send, _, from, to, message}, timestamp}, trace_events) do
+    new_event = {:send, timestamp, from, to, message}
```
jonatanklosko, line 31: If we make it a map, it will likely make the matching cleaner later on, wdyt?

lib/kino/process/tracer.ex
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  # ---- Client API ----
+
+  @doc false
+  def get_trace_events(tracer) do
+    GenServer.call(tracer, :get_trace_events)
+  end
+
+  # ---- Callbacks ----
```
jonatanklosko, line 17:
```suggestion
  def get_trace_events(tracer) do
    GenServer.call(tracer, :get_trace_events)
  end
```

jonatanklosko (commented):
@akoutmos fantastic!!

akoutmos:
> @akoutmos fantastic!!

Thanks @jonatanklosko!

akoutmos:
> Hi @akoutmos, let's please change this one to return `{res, kino}`. Then we will add render_app_tree, render_sup_tree and render_seq_trace that render as a side-effect. The first two will return `Kino.nothing`, the last one will return the result of the function. WDYT? Can you please push those final changes? :)

That sounds good to me :). I think I implemented it as you described in [9a5b965](https://github.com/livebook-dev/kino/pull/165/commits/9a5b965cef23b8b89a939945a1cf42b4e4523ede). Let me know what you think!

josevalim:

akoutmos:
Thanks for the review @josevalim and @jonatanklosko! Happy to see this merged in :)

PR #172, opened 2022-07-27
Update the assert_smart_cell_update function

TylerPachal:
Currently, the `assert_smart_cell_update` performs an exact match against the source of the Kino.  This can be problematic for two reasons:

1. If the source code of your Kino is very large, and you are just asserting on a small change, you have to specify the whole source code and the small change can be hard to spot.

2. If you are testing a smart cell which uses `Kino.SmartCell.prefixed_var_name/2` to create a variable, it is difficult to specify the whole source code because you do not know that variable's name.

---

With this change, there is now an option to pass an arity-one function instead of the complete source as a string.  This allows for custom matching.  I tested it in my own project like this:

```elixir
test "renders code" do
  {kino, source} = start_smart_cell!(MyCell, %{})

  # I can already do a partial match against the source here, which is great.  I just want to make sure this appears somewhere in the source code.  
  assert source =~ "original_file.json"

  push_event(kino, "update_field", %{"field" => "filename", "value" => "updated_file.json"})
  assert_broadcast_event(kino, "update", %{"fields" => %{"filename" => "updated_file.json"}})

  # This is the new part, where I want to do a partial match against the updated source code
  assert_smart_cell_update(kino, %{"filename" => "updated_file.json"}, fn source ->
    source =~ "updated_file.json"
  end)
end
```

---

Let me know what you think!  I have only been using Kino for a few days, it is possible that there is already a better way to do this.

### DIFF
--- a/lib/kino/test.ex
+++ b/lib/kino/test.ex
@@ -115,12 +115,33 @@ defmodule Kino.Test do
   Matches against the source and attribute that are reported as part
   of the update.
 
+  The `source` argument can be a string, or a function.  If it is a
+  string, an exact match is performed.  Pass a function instead
+  if you want to do a custom (non-exact-match) assertion the Kino's
+  source.  The function has one argument, the Kino's source, and is
+  considered successful is the return value is truthy.
+
   ## Examples
 
       assert_smart_cell_update(kino, %{"variable" => "x", "number" => 10}, "x = 10")
 
+      assert_smart_cell_update(kino, %{"status" => "ok"}, fn source -> source =~ "ok" end)
+
   '''
-  defmacro assert_smart_cell_update(kino, attrs, source, timeout \\ 100) do
+  defmacro assert_smart_cell_update(kino, attrs, source, timeout \\ 100)
+
+  defmacro assert_smart_cell_update(kino, attrs, {:fn, _, _} = source_fun, timeout) do
+    quote do
+      %{ref: ref} = unquote(kino)
+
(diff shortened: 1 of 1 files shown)


lib/kino/test.ex
```diff
@@ -115,12 +115,33 @@ defmodule Kino.Test do
...
+  defmacro assert_smart_cell_update(kino, attrs, source, timeout \\ 100)
+
+  defmacro assert_smart_cell_update(kino, attrs, {:fn, _, _} = source_fun, timeout) do
+    quote do
+      %{ref: ref} = unquote(kino)
+
+      assert_receive {:runtime_smart_cell_update, ^ref, unquote(attrs), kino_source, _info},
+                     unquote(timeout)
+
+      assert unquote(source_fun).(kino_source)
```
TylerPachal, line 140: Annotation: This is the main change.  We invoke the function to assert against the `kino_source`, instead of doing an exact (pattern) match in the `assert_receive`

TylerPachal:
Ah, I didn't realize that I could pass a variable through to bind on the underlying `assert_receive`:

```
assert_smart_cell_update(kino, %{"filename" => "updated_file.json"}, updated_source)
assert updated_source =~ "updated_file.json"
```

I will close this PR and open a different one to just update the docs with an example of that.

jonatanklosko:
@TylerPachal exactly, an update to the docs would be appreciated! :cat:

PR #173, opened 2022-07-27
Update docs for assert_smart_cell_update

TylerPachal:
This PR adds documentation about how to get a reference to a Kino's source in `assert_smart_cell_update` without having to perform an exact match on the entire source.

There is a little bit of context in [this PR](https://github.com/livebook-dev/kino/pull/172) I had opened, but it turned out that functionality was already possible but not obvious to me 😛, and not documented.


jonatanklosko (approved):
Thank you! :)

Issue #19, opened 2021-06-17
Add a Kino.Download widget


josevalim:
The goal is to have a button that, when the user clicks, it downloads a file with dynamically generated content, such as: `Kino.Download.start("filename.json", fn -> file_contents end)`. We can get the content type from the filename.

jonatanklosko:
Closed in #174.

Issue #171, opened 2022-07-25
Add support for messaging a single JS view client


jonatanklosko:
Currently we have `broadcast_event` that pushes to everyone, but in certain cases we may want to talk to a specific client. See https://github.com/livebook-dev/livebook/discussions/1297#discussioncomment-3225199.

PR #174, opened 2022-07-27
Add Kino.Download

TylerPachal:
Attempts to address https://github.com/livebook-dev/kino/issues/19.

I am new to Kino, and not a Javascript exeprt, so I would appreciate some feedback on this.

The Javascript code was mostly taken from [this jsfiddle](https://jsfiddle.net/wtp5a0o8/1/).

I tested it in my local livebook instance like this:

Screen Shot 2022-07-27 at 5 13 58 PM

### DIFF
--- a/lib/kino/download.ex
+++ b/lib/kino/download.ex
@@ -1,5 +1,5 @@
 defmodule Kino.Download do
-  @moduledoc ~S'''
+  @moduledoc """
   Built on top of `Kino.JS`, uses Javascript to initiate a download
   on the client's machine.
   '''
--- a/lib/kino/download.ex
+++ b/lib/kino/download.ex
@@ -36,7 +36,7 @@ defmodule Kino.Download do
     """
     export function init(ctx, data) {
       var hiddenElement = document.createElement('a');
-      hiddenElement.href = 'data:attachment/text,' + encodeURI(data.content);
+      hiddenElement.href = 'data:attachment,' + encodeURI(data.content);
       hiddenElement.target = '_blank';
       hiddenElement.download = data.filename;
       hiddenElement.click();
--- a/lib/kino/download.ex
+++ b/lib/kino/download.ex
@@ -1,28 +1,51 @@
 defmodule Kino.Download do
-  @moduledoc false
-
-  use Kino.JS
-  use Kino.JS.Live
-
-  @doc """
-  Renders a download button.  The content of downloaded file is
-  prepared on the server by invoking the `content_fun` argument only
-  when the button is clicked.
+  @moduledoc """
+  A kino for downloading file content.
 
   ## Examples
 
-      Kino.Download.new("file.json", fn ->
-        ~s/{"foo": "bar"}/
+      Kino.Download.new(fn ->
+        "Example text"
       end)
 
+      Kino.Download.new(
+        fn -> Jason.encode!(%{"foo" => "bar"}) end,
+        filename: "data.json"
+      )
+
+      Kino.Download.new(
+        fn -> <<0, 1>> end,
+        filename: "data.bin"
+      )
+
   """
-  def new(filename, content_fun) do
-    Kino.JS.Live.new(__MODULE__, {filename, content_fun})
+
+  use Kino.JS, assets_path: "lib/assets/download"
+  use Kino.JS.Live
+
+  @type t :: Kino.JS.Live.t()
+
+  @doc """
+  Creates a button for file download.
+
+  The given function is invoked to generate the file content whenever
+  a download is requested.
+
+  ## Options
+
+    * `:filename` - the default filename suggested for download
+
+  """
+  @spec new((() -> binary()), keyword()) :: t()
+  def new(content_fun, opts \\ []) do
+    opts = Keyword.validate!(opts, [:filename])
+    filename = opts[:filename]
+    Kino.JS.Live.new(__MODULE__, {content_fun, filename})
   end
 
   @impl true
-  def init({filename, content_fun}, ctx) do
-    {:ok, assign(ctx, filename: filename, content_fun: content_fun)}
+  def init({content_fun, filename}, ctx) do
+    {:ok, assign(ctx, content_fun: content_fun, filename: filename)}
   end
 
   @impl true
@@ -33,50 +56,8 @@ defmodule Kino.Download do
   @impl true
   def handle_event("download", %{}, ctx) do
     file_content = ctx.assigns.content_fun.()
-
     reply_payload = {:binary, %{}, file_content}
-
-    broadcast_event(ctx, "download_content", reply_payload)
-
+    send_event(ctx, ctx.origin, "download_content", reply_payload)
     {:noreply, ctx}
-  end
-
-  asset "main.js" do
-    """
-    export function init(ctx, data) {
-
-      // Handle the event which initiates the actual downloading of the file
-      ctx.handleEvent("download_content", ([info, arrayBuffer]) => {
-        const content = bufferToBase64(arrayBuffer);
-
-        const a = document.createElement('a');
-        a.href = `data:application/octet-stream;base64,${content}`;
-        a.download = data.filename;
-        a.click();
-      });
-
-      // Create the Download button
-      const button = document.createElement('button');
-      button.innerHTML = `Download ${data.filename}`;
-      button.addEventListener('click', buttonClicked);
-      ctx.root.appendChild(button);
-
-      function buttonClicked() {
-        ctx.pushEvent("download", {})
-      }
-    }
-
-    function bufferToBase64(buffer) {
-      let binaryString = "";
-      const bytes = new Uint8Array(buffer);
-      const length = bytes.byteLength;
-
-      for (let i = 0; i < length; i++) {
-        binaryString += String.fromCharCode(bytes[i]);
-      }
-
-      return btoa(binaryString);
-    }
-    """
   end
 end
--- a/lib/kino/download.ex
+++ b/lib/kino/download.ex
@@ -39,7 +39,7 @@ defmodule Kino.Download do
       ctx.assigns.content_fun.()
       |> Base.encode64()
 
-    reply_payload = {:binary, %{filename: ctx.assigns.filename}, file_content}
+    reply_payload = {:binary, %{}, file_content}
 
     broadcast_event(ctx, "do_download", reply_payload)
 
--- a/lib/kino/download.ex
+++ b/lib/kino/download.ex
@@ -51,12 +51,10 @@ defmodule Kino.Download do
         const decoder = new TextDecoder("utf-8");
         const content = decoder.decode(arrayBuffer);
 
-        // Create a hidden element that we can attach the download to
-        const hiddenElement = document.createElement('a');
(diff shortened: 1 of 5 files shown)


lib/kino/download.ex
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    Kino.JS.new(__MODULE__, data)
+  end
+
+  asset "main.js" do
+    """
+    export function init(ctx, data) {
+      var hiddenElement = document.createElement('a');
+      hiddenElement.href = 'data:attachment/text,' + encodeURI(data.content);
```
TylerPachal, line 39: Annotation: For now I hardcoded the type to `attachment/text`.  I tried manually switching it to other values and it never seemed to make any difference, the download always still worked.

test/kino/download_test.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Kino.DownloadTest do
+  use ExUnit.Case, async: true
+
+  describe "start/3" do
+    test "returns %Kino.JS{} " do
```
TylerPachal, line 5: Annotation: I didn't really know what to test here, as-is this test file might not be very useful.

lib/kino/download.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Kino.Download do
+  @moduledoc ~S'''
```
jonatanklosko, line 2:
```suggestion
  @moduledoc """
```

lib/kino/download.ex
```diff
@@ -0,0 +1,46 @@
...
+      Kino.Download.start("file.json", fn -> json end)
+
+  Lazily generating a large file:
+
+      Kino.Download.start("large-file.txt", fn ->
+        String.duplicate("data ", 10_000_000)
+      end)
+  '''
+  @spec start(String.t(), (() -> String.t())) :: Kino.JS.t()
+  def start(filename, content_fun) do
```
jonatanklosko, line 26: We switched the naming convention from `start` to `new` :)

lib/kino/download.ex
```diff
@@ -0,0 +1,46 @@
...
+    data = %{
+      content: content_fun.(),
```
jonatanklosko, line 28: Here we don't actually benefit from the function, since we evaluate it immediately. Also, we don't want to pass it as "static" data to `Kino.JS`, because this will get passed all around in Livebook.

So what we want is `Kino.JS.Live` that receives an event when the user clicks on download, then generates and sends the payload. For a proof of concept we can use `broadcast_event` to sent the payload, though I think we should add `push_event` as per #171 before shipping this.

Note that can event payloads can be `{:binary, %{}, data}`, this is going to be more efficient, and also enable downloading arbitrary formats :)

test/kino/download_test.exs
jonatanklosko: Yeah, there's not much to test with plain `Kino.JS` components, since the more relevant part is on the JS side. Once we have `Kino.JS.Live`, we can test the messages :)

jonatanklosko (commented):
Hey @TylerPachal, thanks for the PR :) A couple comments, since we need a bit different approach.

lib/kino/download.ex
TylerPachal, line 39: Annotation: I need to make this `attachment/text` dynamic.  My problem is that I did not know how to test it.  I tried testing this with `csv`, `json`, and plain text files and leaving the type as `attachment/text` never seemed to break anything.

jonatanklosko: To make it more generic, we can receive any binary from the server, then encode it as base64 and use `data:application/octet-stream;base64;${dataBase64}`. If we need to customize the content type explicitly, we can make it an option in Elixir, but I think we should be fine with `application/octet-stream`, especially that we specify the file name.

lib/kino/download.ex
TylerPachal: Thanks for the detailed feedback!

> an event when the user clicks on download

Do you mean that I should render a button, and attach an event handler to that?  Or is there an event that is already triggered when someone clicks the `Evaluate` button?

jonatanklosko: Yeah, we need a button, on click we will `ctx.pushEvent`, and on the server we will handle that event. Then we also want `ctx.handleEvent` on the client, so the server can push the payload.

jonatanklosko: We don't want to download immediately, because then refreshing a page would trigger the download, we want an explicit action from the user.

TylerPachal: Until we are ready to start the download, where should I store the `content_fun` function?  It is not serialize-able, so I cannot assign it to the context:

```
def init({filename, content_fun}, ctx) do
  # Raises an error:
  # Failed to serialize initial widget data, value #Function<45.65746770/0 in :erl_eval.expr/5> is not JSON serializable, use another data type
  {:ok, assign(ctx, filename: filename, content_fun: content_fun)} 
end
```

jonatanklosko: This context lives only on Elixir side, so you can keep it in assigns with no problem.

I'm guessing your `handle_connect` returns the context or assigns as the payload, it rather return `{:ok, %{}, ctx}`.

TylerPachal: Ah you were right about that, thanks!

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+
+        // Create a hidden element that we can attach the download to
+        const hiddenElement = document.createElement('a');
+        hiddenElement.href = `data:attachment/application/octet-stream;base64,${content}`;
+        hiddenElement.target = '_blank';
+        hiddenElement.download = info.filename;
+        hiddenElement.click();
+      });
+
+      // Create the Download button
```
TylerPachal, line 67: Annotation: Not sure if there is any styling we can add 😅

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+        ~s/{"foo": "bar"}/
+      end)
+
+  """
+  def new(filename, content_fun) do
+    Kino.JS.Live.new(__MODULE__, {filename, content_fun})
+  end
+
+  @impl true
+  @spec init({any, any}, Kino.JS.Live.Context.t()) :: {:ok, Kino.JS.Live.Context.t()}
```
jonatanklosko, line 24: We don't need typespecs for behaviour callbacks, because those already have a typespec :)

```suggestion
```

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+  @impl true
+  def handle_event("prepare_download", %{"filename" => filename}, ctx) do
+    ctx = assign(ctx, filename: filename)
+
+    # To keep things generic the content is a base64-encoded binary
+    file_content =
+      ctx.assigns.content_fun.()
+      |> Base.encode64()
+
+    reply_payload = {:binary, %{filename: ctx.assigns.filename}, file_content}
```
jonatanklosko, line 43: We already send this in initial payload :)

```suggestion
    reply_payload = {:binary, %{}, file_content}
```

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  @impl true
+  @spec init({any, any}, Kino.JS.Live.Context.t()) :: {:ok, Kino.JS.Live.Context.t()}
+  def init({filename, content_fun}, ctx) do
+    {:ok, assign(ctx, filename: filename, content_fun: content_fun)}
+  end
+
+  @impl true
+  def handle_connect(ctx) do
+    {:ok, Map.take(ctx.assigns, [:filename]), ctx}
```
jonatanklosko, line 31:
```suggestion
    {:ok, %{filename: ctx.assigns.filename}, ctx}
```

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  @impl true
+  def handle_event("prepare_download", %{"filename" => filename}, ctx) do
+    ctx = assign(ctx, filename: filename)
+
+    # To keep things generic the content is a base64-encoded binary
+    file_content =
+      ctx.assigns.content_fun.()
+      |> Base.encode64()
```
jonatanklosko, line 41:
```suggestion
    file_content = ctx.assigns.content_fun.()
```

We don't want to encode on Elixir side, because then the payload becomes bigger. With {:binary, %{}, file_content}` we can send an arbitrary Elixir binary, and it is received as ArrayBuffer in JS. You can then convert it to base64, as we do [here](https://github.com/jonatanklosko/kino_lab/blob/450c02fe34549637c2c87b5de46fe196a77832cf/lib/assets/camera_processor/main.js#L119-L129).

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  @impl true
+  def handle_connect(ctx) do
+    {:ok, Map.take(ctx.assigns, [:filename]), ctx}
+  end
+
+  @impl true
+  def handle_event("prepare_download", %{"filename" => filename}, ctx) do
+    ctx = assign(ctx, filename: filename)
```
jonatanklosko, line 36:
```suggestion
  def handle_event("download", %{}, ctx) do
```

We already have the filename on the server :)

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+
+    # To keep things generic the content is a base64-encoded binary
+    file_content =
+      ctx.assigns.content_fun.()
+      |> Base.encode64()
+
+    reply_payload = {:binary, %{filename: ctx.assigns.filename}, file_content}
+
+    broadcast_event(ctx, "do_download", reply_payload)
```
jonatanklosko, line 45:
```suggestion
    broadcast_event(ctx, "download_content", reply_payload)
```

lib/kino/download.ex
```diff
@@ -0,0 +1,79 @@
...
+      ctx.handleEvent("do_download", ([info, arrayBuffer]) => {
+        const decoder = new TextDecoder("utf-8");
+        const content = decoder.decode(arrayBuffer);
+
+        // Create a hidden element that we can attach the download to
+        const hiddenElement = document.createElement('a');
+        hiddenElement.href = `data:attachment/application/octet-stream;base64,${content}`;
+        hiddenElement.target = '_blank';
+        hiddenElement.download = info.filename;
+        hiddenElement.click();
```
jonatanklosko, line 64:
```suggestion
        const a = document.createElement('a');
        a.href = `data:application/octet-stream;base64,${content}`;
        a.download = info.filename;
        a.click();
```

Also worth extracting into a separate function.

lib/kino/download.ex
jonatanklosko: Yeah, we can use any CSS we want, but don't worry about it, I can add some Livebook-matching style later :)

TylerPachal:
Thanks for your help & feedback Jonatan!  I am going to leave this here for now while we wait on https://github.com/livebook-dev/kino/issues/171

jonatanklosko:
Sounds good, thanks :)

jonatanklosko:
We have `send_event/4` now, so I made final adjustments :)

image

@josevalim wdyt?

josevalim:
We should also add a label option. The text that we show should be: `opts[:label] || opts[:filename] || "Download"`. other than that, it looks great to me!

PR #176, opened 2022-07-29
Fix Mermaid error when inspected process name starts with colon

mat-hek:
When using `Kino.Process.app_tree` when some process was registered under a name like `:"some process"`, I got a Mermaid error like this:

```
Mermaid
Parse error on line 4:
...ne):::root ---> 1(:"some process...
-----------------------^
Expecting 'SEMI', 'NEWLINE', 'SPACE', 'EOF', 'GRAPH', 'DIR', 'subgraph', 'SQS', 'SQE', 'end', 'AMP', 'DOUBLECIRCLEEND', 'PE', '-)', 'STADIUMEND', 'SUBROUTINEEND', 'ALPHA', 'COLON', 'PIPE', 'CYLINDEREND', 'DIAMOND_STOP', 'TAGEND', 'TRAPEND', 'INVTRAPEND', 'START_LINK', 'LINK', 'STYLE', 'LINKSTYLE', 'CLASSDEF', 'CLASS', 'CLICK', 'DOWN', 'UP', 'DEFAULT', 'NUM', 'COMMA', 'MINUS', 'BRKT', 'DOT', 'PCT', 'TAGSTART', 'PUNCTUATION', 'UNICODE_TEXT', 'PLUS', 'EQUALS', 'MULT', 'UNDERSCORE', got 'STR'
```

Figured out that's because of the colon at the beginning of the inspected name, so this fix removes the colon if present. Possibly there should be more proper escaping, but that worked for me ;)

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -576,7 +576,7 @@ defmodule Kino.Process do
 
     "#{id}(#{display}):::#{type}"
   end
-  
+
   defp module_or_atom_to_string(atom) do
     case Atom.to_string(atom) do
       "Elixir." <> rest -> rest
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -573,7 +573,7 @@ defmodule Kino.Process do
       |> Process.info(:registered_name)
       |> case do
         {:registered_name, []} -> inspect(pid)
-        {:registered_name, name} -> inspect(name)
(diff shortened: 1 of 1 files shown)


josevalim:
@mat-hek I pushed a simpler approach, can you please confirm it works?

mat-hek:
@josevalim Works for me. Won't work for things like `:":\"name\""`, but maybe that's too weird to support.

josevalim:

PR #179, opened 2022-08-03
Add layout utilities

### LINKED ISSUE #157: Grid of Frames
It'd be great if there was a way to show outputs side-by-side natively somehow.
The easiest way I can think of right now would be a `Grid` or `Row` that exposes several frames that can be rendered into.
The actual display would then show these frames side-by-side.

I really like the Matrix example from "Plotting with VegaLite" and think having a way to do this with arbitrary component would be great
For reference, said example:
image

(From what I can tell there is no way to implement this externally, but please correct me if there is a way)
(I'd be happy to try and contribute something here if desired)

### LINKED ISSUE #164: Introduce Kino.Tab
We should change some components to render multiple tabs by default. Like a PID/atom with supervision tree and so forth.

jonatanklosko:
Closes #164, closes #157.

[kino_layout.webm](https://user-images.githubusercontent.com/17034772/182672612-c3ffd028-6f15-43c5-9612-59171e359520.webm)

### DIFF
--- a/lib/kino/layout.ex
+++ b/lib/kino/layout.ex
@@ -22,8 +22,8 @@ defmodule Kino.Layout do
       ]
 
       Kino.Layout.tabs([
-        {"Table", Kino.DataTable.new(data)},
-        {"Raw", data}
+        Table: Kino.DataTable.new(data),
+        Raw: data
       ])
 
   """
--- a/lib/kino/layout.ex
+++ b/lib/kino/layout.ex
@@ -1,0 +1,70 @@
+defmodule Kino.Layout do
+  @moduledoc """
+  Layout utilities for arranging multiple kinos together.
+  """
+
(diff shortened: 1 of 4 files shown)


lib/kino/layout.ex
```diff
@@ -0,0 +1,70 @@
...
+  ## Examples
+
+      data = [
+        %{id: 1, name: "Elixir", website: "https://elixir-lang.org"},
+        %{id: 2, name: "Erlang", website: "https://www.erlang.org"}
+      ]
+
+      Kino.Layout.tabs([
+        {"Table", Kino.DataTable.new(data)},
+        {"Raw", data}
```
josevalim, line 26: Perhaps?

```suggestion
        Table: Kino.DataTable.new(data),
        Raw: data
```

josevalim (approved):

PR #183, opened 2022-08-04
Add send_event/4 for Kino.JS.Live

### LINKED ISSUE #171: Add support for messaging a single JS view client

jonatanklosko:
Closes #171.

[send_event.webm](https://user-images.githubusercontent.com/17034772/182946178-99d2b0ba-01b6-47db-ae92-eab9aa4fb2c7.webm)

Note that if the same JS view is rendered multiple times, all instances are still the same client, because client.

### DIFF
--- a/lib/kino/js/live/server.ex
+++ b/lib/kino/js/live/server.ex
@@ -26,13 +26,10 @@ defmodule Kino.JS.Live.Server do
   def send_event(ctx, origin, event, payload) do
     ref = ctx.__private__.ref
 
-    pid =
-      case ctx.__private__.clients[origin] do
-        {pid, _} -> pid
-        _ -> raise "could not find a connected client with origin #{inspect(origin)}"
-      end
+    with {pid, _} <- ctx.__private__.clients[origin] do
+      Kino.Bridge.send(pid, {:event, event, payload, %{ref: ref}})
+    end
 
-    Kino.Bridge.send(pid, {:event, event, payload, %{ref: ref}})
     :ok
   end
 
--- a/lib/kino/js/live/server.ex
+++ b/lib/kino/js/live/server.ex
@@ -23,6 +23,19 @@ defmodule Kino.JS.Live.Server do
     :ok
   end
 
+  def send_event(ctx, origin, event, payload) do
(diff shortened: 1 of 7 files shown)


lib/kino/js/live/server.ex
```diff
@@ -23,6 +23,19 @@ defmodule Kino.JS.Live.Server do
     :ok
   end
 
+  def send_event(ctx, origin, event, payload) do
+    ref = ctx.__private__.ref
+
+    pid =
+      case ctx.__private__.clients[origin] do
+        {pid, _} -> pid
+        _ -> raise "could not find a connected client with origin #{inspect(origin)}"
```
josevalim, line 32: I think it may be better to just execute a no-op because you can always have race conditions: i.e. between you computing the message to send and sending it, the client may have been disconnected.

That's why sending processes in Erlang almost always no-op if the process is dead.

lib/kino/js/live/server.ex
```diff
@@ -116,4 +142,28 @@ defmodule Kino.JS.Live.Server do
...
+      ctx
+    else
+      monitor_ref = Kino.Bridge.monitor(pid)
+      put_in(ctx.__private__.clients[origin], {pid, monitor_ref})
+    end
+  end
+
+  defp remove_client(ctx, ref) do
+    origin =
+      Enum.find_value(ctx.__private__.clients, fn
```
josevalim, line 157: It would be best to not do a full traversal. Maybe we could store the origin (or origins) in `__private__.refs`?

jonatanklosko: Ah, my initial thought was that it's fine to do it whenever someone disconnects, but now I've realised we do this on every `:DOWN` messages that the user may be expecting. I will optimise with a separate map.

lib/kino/js/live/server.ex
jonatanklosko: Good call!

josevalim (approved):

PR #185, opened 2022-08-05
Add default tabbed output for pids and atoms

### LINKED ISSUE #182: Add default tabbed output for pids and atoms
If the atom is an application with supervision tree, return a tab with the "Raw" representation as default and the application tree as alternative representation.

If the atom is a named process representing a supervisor, return a tab with the "Raw" representation as default and the supervision tree as alternative representation.

If the pid represents a supervisor, return a tab with the "Raw" representation as default and the supervision tree as alternative representation.

jonatanklosko:
Closes #182.

[tabbed_atom_pid.webm](https://user-images.githubusercontent.com/17034772/183126271-26166836-39a3-4f27-9b2e-98692f47bb65.webm)

### DIFF
--- a/lib/kino/raw.ex
+++ b/lib/kino/raw.ex
@@ -1,0 +1,19 @@
+defmodule Kino.Raw do
+  @moduledoc """
+  A struct wrapping any term for default rendering.
+
+  This is just a meta-struct that implements the `Kino.Render`
+  protocol, so that the wrapped value is rendered using the inspect
+  protocol.
+  """
+
+  defstruct [:term]
+
+  @opaque t :: %__MODULE__{term: term()}
+
+  @doc """
+  Wraps the given term.
+  """
+  @spec new(term()) :: t()
+  def new(term), do: %__MODULE__{term: term}
+end
--- a/lib/kino/render.ex
+++ b/lib/kino/render.ex
@@ -18,6 +18,12 @@ defimpl Kino.Render, for: Any do
   end
 end
 
+defimpl Kino.Render, for: Kino.Raw do
+  def to_livebook(raw) do
+    Kino.Output.inspect(raw.term)
+  end
+end
+
 defimpl Kino.Render, for: Kino.JS do
   def to_livebook(kino) do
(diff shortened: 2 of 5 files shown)


lib/kino/render.ex
```diff
@@ -80,6 +80,20 @@ end
...
+defmodule Kino.Raw do
+  @moduledoc false
+
+  defstruct [:term]
+
+  def new(term), do: %__MODULE__{term: term}
+
+  defimpl Kino.Render do
+    def to_livebook(raw) do
+      Kino.Output.inspect(raw.term)
```
josevalim, line 92: It is probably more useful to have a `Kino.Text` that is rendered as is (and then we pass the inspected representation) rather than Kino.Raw that is inspect based?

josevalim (approved):

lib/kino/render.ex
jonatanklosko: I don't follow, you mean calling `inspect` explicitly? That would just be more work because we would need to duplicate the inspect logic. The only reason for `Kino.Raw` is to avoid infinite loop.

josevalim: That's a good point. Here is the issue though: Axon is going to run into the same issue once they implement tabbed outputs. Should we tell them to use Kino.Raw?

jonatanklosko: Ah totally, I made `Kino.Raw` public and we can use it in those cases.

lib/kino/raw.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Kino.Raw do
```
josevalim, line 1: Should we call it `Kino.Inspect` then?

I think it would go better with the description.

josevalim: Also, should we make it so Kino.inspect uses this? :D

jonatanklosko: `Kino.Inspect` sounds good! `Kino.inspect` handles optional label, so I'd keep it as is.

josevalim (approved):

PR #186, opened 2022-08-06
Add Kino.listen and expand Kino.animate

### LINKED ISSUE #168: Kino.Control.stream/3 should expect the function to return either {:cont, val} or :halt
We will have to deprecate incompatible return types.

### LINKED ISSUE #177: Simpler interface for `Kino.animate/3`
Right now, the interface for [Kino.animate/3](https://hexdocs.pm/kino/Kino.html#animate/3) is:

```elixir
Kino.animate(100, 0, fn i ->
  md = Kino.Markdown.new("**Iteration: `#{i}`**")
  {:cont, md, i + 1}
end)
```

Would there be an interest in providing an alternative interface using ranges (or any collection)? I.e.

```elixir
Kino.animate(1..10, fn i -> Kino.Markdown.new("**Iteration: `#{i}`**") end)
```

Configuration could be done with options such as `:interval` (with the default being `1000`) and `:repeat` (with the default being `true`). I.e.

```elixir
Kino.animate(1..10, fn i -> Kino.Markdown.new("**Iteration: `#{i}`**") end, interval: 100, repeat: false)
```

I'm happy to implement this if there's an interest!

jonatanklosko:
Closes #168, closes #177.

[`Kino.listen/2,3`](https://youtu.be/RmGe-LY5HQs?t=8) replaces `Kino.Control.stream/2,3` and is more general because it accepts any stream. Similarly `Kino.animate/2` is added and `Kino.animate/3` is extended to handle streams.

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -167,37 +167,93 @@ defmodule Kino do
   Renders a kino that periodically calls the given function
   to render a new result.
 
-  The callback is run every `interval_ms` milliseconds and receives
-  the accumulated value. The callback should return either of:
-
-    * `{:cont, term_to_render, acc}` - the continue
-
-    * `:halt` - to no longer schedule callback evaluation
+  The callback receives a stream element and should return a term
+  to be rendered.
 
   This function uses `Kino.Frame` as the underlying kino.
   It returns nothing (a non-printable result).
 
   ## Examples
 
+  An animation is created by turning a stream of values into
+  subsequent animation frames:
+
+      Stream.interval(100)
+      |> Stream.take(100)
+      |> Kino.animate(fn i ->
+        Kino.Markdown.new("**Iteration: `#{i}`**")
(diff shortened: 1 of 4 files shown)


lib/kino.ex
```diff
@@ -167,37 +167,93 @@ defmodule Kino do
...
+              "please use Kino.animate/2 or pass an arity-2 function"
+          )
+
+          fn _i, state -> fun.(state) end
+
+        is_function(fun, 2) ->
+          fun
+      end
+
+    Kino.Frame.periodically(frame, interval_ms, {0, state}, fn {i, state} ->
```
jonatanklosko, line 252: We could just call `animate(Stream.interval(interval_ms), state, fun)` instead, but in this specific case we can re-use the frame process, so it's probably worth it.

josevalim (approved):

PR #191, opened 2022-08-25
Add a custom backend for Kernel.dbg/2

jonatanklosko:
Registers a custom backend for `Kernel.dbg/2`. For pipelines it renders an interactive editor, otherwise it renders the result with the `Kino.Render` protocol.

[kino_dbg.webm](https://user-images.githubusercontent.com/17034772/187455667-b166ce50-c440-444c-94dc-e2c0280a4924.webm)


lib/kino/js/live.ex
```diff
@@ -194,6 +194,45 @@ defmodule Kino.JS.Live do
   Invoked when the server is started.
 
   See `c:GenServer.init/1` for more details.
+
+  > #### Starting other kinos {: .warning}
+  >
+  > It is generally not possible to start kinos inside the `c:init/2`
+  > callback, as such operation would block forever. In case you need
+  > to start other kinos during initialization, you mast start them
```
hugobarauna, line 202:
```suggestion
  > to start other kinos during initialization, you must start them
```

lib/kino/bridge.ex
```diff
@@ -120,6 +120,17 @@ defmodule Kino.Bridge do
...
   end
 
+  @doc """
+  Returns the file that is currently being evaluated.
+  """
+  @spec get_evaluation_file() :: String.t()
+  def get_evaluation_file() do
+    case io_request(:livebook_get_evaluation_file) do
+      {:ok, file} -> file
+      {:error, _} -> "unknown"
```
josevalim, line 130:
```suggestion
      {:error, _} -> "nofile"
```

That's the default "i don't know the file" in Elixir.

lib/kino/debug.ex
```diff
@@ -0,0 +1,125 @@
...
+        arg = Macro.var(:arg, __MODULE__)
+        expr = Macro.pipe(arg, ast, 0)
+
+        quote do
+          fn unquote(arg) -> unquote(expr) end
+        end
+      end
+
+    quote do
+      unquote(assignments)
```
josevalim, line 69: I don't think we need the assignments? It seems we only care about `head_var` and the result (the last var). In this case, I think we only need this:

```elixir
head_var = Macro.generate_unique_argument(:arg, __MODULE__)
result = Enum.reduce(rest_asts, head_var, &Macro.pipe(& &2, & &1, 0))

quote do
  unquote(head_var) = unquote(head_ast)
  unquote(__MODULE__).__dbg__(...)
  unquote(result)
end
```

josevalim: Furthermore, to avoid introducing the same AST multiple times on the dbg callback, instead of an if/else at the root, I think you actually want to move the `pid` handling here. So overall you would have this:

```
head_var = Macro.generate_unique_argument(:arg, __MODULE__)
result = Enum.reduce(rest_asts, head_var, &Macro.pipe(& &2, & &1, 0))

quote do
  unquote(head_var) = unquote(head_ast)

  unless lookup_pid(...) do
    wrapped_funs = fn -> unquote(funs) end
    unquote(__MODULE__).__dbg__(sources, head_var, ...)
  end

  unquote(result)
end
```

lib/kino/debug.ex
```diff
@@ -0,0 +1,125 @@
...
+  """
+  @spec dbg(Macro.t(), Macro.t(), Macro.Env.t()) :: Macro.t()
+  def dbg(ast, options, %Macro.Env{} = env) do
+    default_dbg_ast = Macro.dbg(ast, options, env)
+    dbg_id = System.unique_integer()
+
+    case ast do
+      {:|>, _meta, _args} ->
+        quote do
+          if pid = unquote(__MODULE__).lookup_dbg_handler(unquote(dbg_id)) do
```
josevalim, line 21: Quick question: what happens if `dbg` was invoked from a separate process? Do we also want to show the "fancy" version?

josevalim:
Btw, ehe "Copy new pipeline" text should also be clickable :)

It would also be great if we could use keyboard navigation once a pipe is focused!

josevalim:
Finally, after playing with it for a little bit, I think we should do some final changes to the selector. Take this example:

image

If I drag the fourth item to the second position, then the new fourth item should be

lib/kino/debug.ex
jonatanklosko: @josevalim currently we pass all intermediate results to the kino, so for the first render we don't need to run all the functions. Do you think it's better to only pass the initial value and reevaluate?

lib/kino/debug.ex
jonatanklosko: Actually that's an interesting case, I would say there is no reason to differentiate that, however the relevant detail is that the started kino is tied to the process lifetime, so when debugging a process that crashes this is an issue.

jonatanklosko:
@josevalim there are cases where tracking selection on row basis is also unintuitive. For example:

image

When we move filter above shuffle, it makes more sense to keep map selected to see what impact filter had, rather than select shuffle.

I actually think always selecting the last one is less arbitrary than that, so it's fine by me :)

lib/kino/debug.ex
josevalim: Hrm… this is tricky because the only times they will be different are when using random or when it depends on something from the current process.

it is probably best to show the original results then, so please ignore this particular bit. :) I would still move the pid lookup inside!

lib/assets/debug_pipeline/main.css
```diff
@@ -0,0 +1,203 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+.navigation {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 12px;
```
josevalim, line 31: With the smaller font, the original 8px are the correct value for equal spacing on both top and bottom. :)

```suggestion
  margin-bottom: 8px;
```

josevalim (approved):

jonatanklosko:
FTR I added tests and a demo to the PR description :)

PR #205, opened 2022-11-07
Use `Macro.classify_atom/1`

aleDsz:


josevalim (approved):

jonatanklosko (approved):
Looks good to me, but I would keep this open until we are ready to release 0.8, otherwise it's a breaking change. @josevalim does that make sense?

josevalim:
Sounds good to keep it open for now!

PR #208, opened 2022-11-15
Add Kino.Tree kino

### LINKED ISSUE #203: Exploring nested data visually in Livebook
Hi, it would be great if Livebook had a feature that would allow exploring nested data visually (something akin of a tree view component). My main use case for this is exploring RESTful APIs which sometimes throw massive amounts of JSON at you and you just want to get the feel for the structure of the response (I think this could be a step towards Livebook becoming a simple alternative to tools like [Postman](https://www.postman.com/)).

I built a prototype of this as [part of the recent SpawnFest](https://github.com/spawnfest/the-arborists). Here's what I have in mind:

demo-small

The approach I took is somewhere between extending the inspect protocol and a custom Kino. In my implementation, `Kino.Tree` is not really a generic tree Kino (as compared to `Kino.DataTable`/`Kino.Table`) since it contains specific features: e.g. the item text changes based on whether it's collapsed or expanded; the tree adds/skips trailing commas. On the other hand, `Kino.Tree` is not exactly on par with what happens when you evaluate Elixir expressions (no colors, ignores inspection limits, kind of reimplements the inspect protocol). It would be great to somehow reconcile the differences between the two. Maybe this should be implemented in Livebook directly? An argument for having this as a Kino is that we can add more features, like UI controls to control the folding level (e.g. fold all levels below 3).

I'd love to hear your thoughts and prepare a PR for this.

stefanchrobot:
Closes #203.

### DIFF
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -1,0 +1,38 @@
+.app {
+  font-family: "JetBrains Mono", monospace;
+  font-size: 0.875rem;
+}
+
+div.item {
+  display: flex;
+}
+
+div.clickable {
+  cursor: pointer;
+}
+
+.ri {
+  font-size: 1.25rem;
+  vertical-align: middle;
+}
+
+.ri-empty {
+  display: inline-block;
+  width: 1.25rem;
+}
+
+ol {
+  display: none;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  padding-left: 2rem;
+}
+
+ol.show {
+  display: block;
+}
+
+div.suffix {
+  margin-left: 1.25rem;
+}
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -67,6 +67,15 @@ ol.show {
   display: block;
 }
 
+ol li {
+  display: flex;
+  flex-direction: column;
+}
+
 div.suffix {
   margin-left: 2ch;
+}
+
+span.code {
+  white-space: pre;
 }
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -22,6 +22,12 @@ div.clickable {
   cursor: pointer;
 }
 
+div.icon-container {
+  display: inline-block;
+  width: 2ch;
+  margin-right: 2px;
+}
+
 .ri {
   font-size: 1.25rem;
   vertical-align: middle;
@@ -29,7 +35,7 @@ div.clickable {
 
 .ri-empty {
   display: inline-block;
-  width: 1.25rem;
+  width: 2ch;
 }
 
 ol {
@@ -37,7 +43,7 @@ ol {
   list-style-type: none;
   margin: 0;
   padding: 0;
-  padding-left: 2rem;
+  margin-left: 2ch;
 }
 
 ol.show {
@@ -45,5 +51,5 @@ ol.show {
 }
 
 div.suffix {
-  margin-left: 1.25rem;
+  margin-left: 2ch;
 }
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -40,6 +40,7 @@ div.clickable {
 }
 
 div.icon-container {
+  flex-shrink: 0;
   display: inline-block;
   width: 2ch;
   margin-right: 2px;
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -1,6 +1,17 @@
 .app {
   font-family: "JetBrains Mono", monospace;
   font-size: 0.875rem;
+
+  --gray-50: #f8fafc;
+  --gray-100: #f0f5f9;
+  --gray-200: #e1e8f0;
+  --gray-300: #cad5e0;
+  --gray-400: #91a4b7;
+  --gray-500: #61758a;
+  --gray-700: #304254;
+  --gray-800: #1c2a3a;
+
+  color: var(--gray-500);
 }
 
 div.item {
--- a/lib/assets/tree/main.css
+++ b/lib/assets/tree/main.css
@@ -11,6 +11,23 @@
   --gray-700: #304254;
   --gray-800: #1c2a3a;
 
+  --ansi-color-black: black;
+  --ansi-color-red: #ca1243;
+  --ansi-color-green: #50a14f;
+  --ansi-color-yellow: #c18401;
+  --ansi-color-blue: #4078f2;
+  --ansi-color-magenta: #a726a4;
+  --ansi-color-cyan: #0184bc;
+  --ansi-color-white: white;
+  --ansi-color-light-black: #5c6370;
+  --ansi-color-light-red: #e45649;
+  --ansi-color-light-green: #34d399;
+  --ansi-color-light-yellow: #fde68a;
+  --ansi-color-light-blue: #61afef;
+  --ansi-color-light-magenta: #c678dd;
+  --ansi-color-light-cyan: #56b6c2;
+  --ansi-color-light-white: white;
+
   color: var(--gray-500);
 }
 
--- a/lib/assets/tree/main.js
+++ b/lib/assets/tree/main.js
@@ -1,22 +1,5 @@
 import * as Vue from "https://cdn.jsdelivr.net/npm/vue@3.2.26/dist/vue.esm-browser.prod.js";
 
-// TODO: use Macro.classify_atom/1 after switching to Elixir 1.14
-
-const SIMPLE_ATOM = /^[a-zA-Z0-9_]+$/;
-const MODULE_NAME = /^Elixir\.[a-zA-Z0-9_]+$/;
-
-function atomToString(atom) {
-  if (MODULE_NAME.test(atom)) {
-    return atom.slice(7);
-  } else if (atom === "nil" || atom === "true" || atom === "false") {
-    return atom;
-  } else if (SIMPLE_ATOM.test(atom)) {
-    return `:${atom}`;
-  } else {
-    return `:"${atom.replaceAll('"', '\\"')}"`;
-  }
-}
-
 export function init(ctx, tree) {
   ctx.importCSS("main.css");
   ctx.importCSS("https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap");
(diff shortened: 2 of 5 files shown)


lib/assets/tree/main.js
```diff
@@ -0,0 +1,154 @@
...
+  if (MODULE_NAME.test(atom)) {
+    return atom.slice(7);
+  } else if (atom === "nil" || atom === "true" || atom === "false") {
+    return atom;
+  } else if (SIMPLE_ATOM.test(atom)) {
+    return `:${atom}`;
+  } else {
+    return `:"${atom.replaceAll('"', '\\"')}"`;
+  }
+}
```
stefanchrobot, line 18: Use `Macro.classify_atom/1` here once Elixir 1.14 is required, similar to https://github.com/livebook-dev/kino/pull/205.

lib/kino/tree.ex
```diff
@@ -0,0 +1,95 @@
...
+      Kino.Tree.new(data)
+  """
+
+  use Kino.JS, assets_path: "lib/assets/tree"
+
+  def new(data) do
+    Kino.JS.new(__MODULE__, to_node(data))
+  end
+
+  def to_node(string) when is_binary(string) do
```
jonatanklosko, line 32: Let's make this private. For the tests you can define a private function like this:

```elixir
defp tree(%Kino.JS{ref: ref}) do
  send(Kino.JS.DataStore, {:connect, self(), %{origin: "client:#{inspect(self())}", ref: ref}})
  assert_receive {:connect_reply, data, %{ref: ^ref}}
  data
end
```

then we can make the same assertions against the received data :)

lib/assets/tree/main.css
```diff
@@ -0,0 +1,38 @@
...
+  display: inline-block;
+  width: 1.25rem;
+}
+
+ol {
+  display: none;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  padding-left: 2rem;
```
jonatanklosko, line 29: Ideally we would make the indentation consistent with arrow-less inspect. Here we can use `2ch`, which corresponds to the width of two spaces.

For the arrow we can't use `2ch` directly because it uses the icon font, but I we can wrap the icon in another element, make that element `2ch` wide and then maybe align the icon if needed.

lib/assets/tree/main.js
jonatanklosko: Any reason we don't `inspect` the atom into string upfront?

lib/assets/tree/main.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+ol.show {
+  display: block;
+}
+
+div.suffix {
+  margin-left: 1.25rem;
+}
```
jonatanklosko, line 38: Ideally we would also color the items to match inspect, but we can do this separately. For now maybe let's use `#61758A` that we use for regular output?

jonatanklosko:
Great stuff! A couple comments :)

lib/kino/tree.ex
stefanchrobot: 🤩 Awesome! Switching to testing over the public API.

lib/assets/tree/main.js
stefanchrobot: Just didn't occur to me that `inspect` does exactly what I want here. But this makes me think we should push all the formatting to the backend. I'll try that approach and let's see what happens.

lib/assets/tree/main.css
stefanchrobot: 👍 Changed.

Where can I find the colors that the inspect is using?

jonatanklosko: You can look at the names [here](https://github.com/livebook-dev/livebook/blob/42d2386a2d4d3d115946fc6d9a98f2df6319f7a4/lib/livebook/runtime/evaluator/default_formatter.ex#L69-L87) and then their colors [here](https://github.com/livebook-dev/livebook/blob/42d2386a2d4d3d115946fc6d9a98f2df6319f7a4/assets/css/ansi.css) :)

lib/assets/tree/main.css
stefanchrobot: 👍 That did the job!

lib/assets/tree/main.css
stefanchrobot: Thanks! I'll add colors to this PR.

lib/assets/tree/main.js
```diff
@@ -0,0 +1,61 @@
...
+        name: "TreeNode",
+        props: ["node", "level", "lastChild"],
+        data() {
+          return {
+            collapsed: this.level !== 1,
+          };
+        },
+        methods: {
+          itemText() {
+            return this.node.children == null || this.collapsed
```
jonatanklosko, line 33:
```suggestion
            return this.node.children === null || this.collapsed
```

lib/assets/tree/main.js
```diff
@@ -0,0 +1,61 @@
...
+              ? this.node.text
+              : this.node.expanded.prefix;
+          },
+          suffix() {
+            return this.node.children == null || this.collapsed
```
jonatanklosko, line 38:
```suggestion
            return this.node.children === null || this.collapsed
```

lib/assets/tree/main.js
```diff
@@ -0,0 +1,61 @@
...
+
+export function init(ctx, tree) {
+  ctx.importCSS("main.css");
+  ctx.importCSS("https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap");
+  ctx.importCSS("https://cdn.jsdelivr.net/npm/remixicon@2.5.0/fonts/remixicon.min.css");
+
+  const app = Vue.createApp({
+    template: `
+      <div class="app">
+        <!-- add controls here -->
```
jonatanklosko, line 11:
```suggestion
```

lib/kino/tree.ex
```diff
@@ -0,0 +1,139 @@
...
+  A kino for interactively viewing nested data as a tree view.
+
+  The data can be any term.
+
+  ## Examples
+
+      data = %{
+        id: 1,
+        email: "user@example.com",
+        inserted_at: ~U[2022-01-01T10:00:00Z],
```
jonatanklosko, line 12: We should specialcase structs like dates and use their inspect, rather than showing a map.

Another option would be to use inspect on structs that implement it (rather than using the default inspect implementation), we would need to make sure multiline items are rendered fine. @josevalim any preference?

lib/kino/tree.ex
```diff
@@ -0,0 +1,139 @@
...
+        ]
+      }
+
+      Kino.Tree.new(data)
+  """
+
+  use Kino.JS, assets_path: "lib/assets/tree"
+
+  def new(data) do
+    Kino.JS.new(__MODULE__, to_node(data))
```
jonatanklosko, line 29: We can put this kino in a grid to get a bordered box as we do with inspect, wdyt?

image

lib/kino/tree.ex
```diff
@@ -0,0 +1,139 @@
...
+    %{text: inspect(other), children: nil}
+  end
+
+  defp to_key_value_node({key, value}) do
+    key_text =
+      if is_atom(key) do
+        String.trim_leading(inspect(key), ":") <> ": "
+      else
+        inspect(key) <> " => "
+      end
```
jonatanklosko, line 105:
```suggestion
      case inspect(key) do
        ":" <> name when is_atom(key) -> name <> ":"
        key -> key <> " => "
      end
```

jonatanklosko (commented):
Beautiful!

lib/kino/tree.ex
stefanchrobot: Nice!

lib/kino/tree.ex
stefanchrobot: Definitely. Looks better to me!

lib/assets/tree/main.js
```diff
@@ -0,0 +1,57 @@
...
+        data() {
+          return { expanded: this.level === 1 };
+        },
+        methods: {
+          style(span) {
+            return span.color == null ? {} : {color: span.color};
+          },
+          nbsp(text) {
+            return text.replaceAll(' ', '\xa0');
+          },
```
jonatanklosko, line 34: Instead of nbsp you can use `white-space: pre` on the element :)

lib/assets/tree/main.css
```diff
@@ -0,0 +1,72 @@
...
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  margin-left: 2ch;
+}
+
+ol.show {
+  display: block;
+}
+
```
jonatanklosko, line 69:
```suggestion

ol li {
  display: flex;
}

```

When the line is too long, the `2ch` indentation element shrinks to 0 width, changing the list item to flex fixes that.

lib/kino/tree.ex
```diff
@@ -0,0 +1,173 @@
...
+        ]
+      }
+
+      Kino.Tree.new(data)
+  """
+
+  use Kino.JS, assets_path: "lib/assets/tree"
+
+  def new(data) do
+    Kino.Layout.grid([Kino.JS.new(__MODULE__, to_node(data, []))], boxed: true)
```
jonatanklosko, line 29:
```suggestion
    tree = to_node(data, [])
    kino = Kino.JS.new(__MODULE__, tree)
    Kino.Layout.grid([kino], boxed: true)
```

lib/kino/tree.ex
```diff
@@ -0,0 +1,173 @@
...
+
+      %{content: content} = node ->
+        %{node | content: [key_span, sep_span | content]}
+    end
+  end
+
+  defp to_children(items, container_size) do
+    items
+    |> Enum.with_index()
+    |> Enum.map(fn {item, index} -> to_node(item, suffix(index, container_size)) end)
```
jonatanklosko, line 151:
```suggestion
    Enum.with_index(items, fn item, index ->
      to_node(item, suffix(index, container_size))
    end)
```

same below :)

test/kino/tree_test.exs
```diff
@@ -0,0 +1,224 @@
...
+    }
+  end
+
+  defp plaintext(%{content: content, children: nil} = node) do
+    %{node | content: text_of(content)}
+  end
+
+  defp text_of(list) when is_list(list) do
+    list |> Enum.map(& &1.text) |> Enum.join()
+  end
```
jonatanklosko, line 45: Let's move all the private functions to the bottom :)

lib/kino/tree.ex
josevalim: I think using inspect for structs that implement it is the safest/best choice. Otherwise we can leak a lot of internals, such as in Nx.Tensor. In other words, check if the struct `Inspect.impl_for(data)` returns `Inspect.Any` or anything else.

lib/assets/tree/main.js
stefanchrobot: 👍

lib/assets/tree/main.css
stefanchrobot: Nice! Changed.

lib/kino/tree.ex
stefanchrobot: Much better!

lib/kino/tree.ex
stefanchrobot: I should have looked into the docs better! I tried `Enum.with_index(items, fn {item, index} -> ... end)` and gave up when that didn't work.

test/kino/tree_test.exs
stefanchrobot: 👍

lib/kino/tree.ex
stefanchrobot: Pushed the commit with the change.

Should we add any color to those items? Livebook's code editor uses cyan, but renders it as grey. Regular expressions use red.

jonatanklosko: I think grey is fine in this case. We just need to fix the spacing for multiline inspect:

```elixir
%{
  a: "first",
  tensor: Nx.iota({100, 100}),
  z: "last"
}
|> Kino.Tree.new()
```

After that we are ready to go!

jonatanklosko: I think I got it, will push in a moment.

jonatanklosko: image

jonatanklosko (approved):
Awesome job!! :shipit:

josevalim (approved):
:heart:

PR #211, opened 2022-11-21
Glide Data Grid - Initial integration

cristineguadelupe:
This PR converts the current table to Glide Data Grid tables and adds the following functionalities:

- Items per page

https://user-images.githubusercontent.com/5660941/202933987-f192779e-9cde-4563-a9c3-db8c1af6e31d.mp4

- Basic `cellKind` inference

https://user-images.githubusercontent.com/5660941/202934091-2d2abce8-d785-48b2-ad4e-83d1737e430b.mp4

- Data Grid defaults: search, resize, selections... (client side)

https://user-images.githubusercontent.com/5660941/202934506-96436d72-81cb-49cf-bd71-b51d61a5083f.mp4

**Known Issues**

- Selections and how to clear the selections are a bit confusing. We can improve this in the future by controlling the selections manually.

- Column selection conflicts with sorting: Both are triggered by `onHeaderClicked`. I don't think this is a big problem, so I kept the column selection active. This can also be solved by manually controlling the selections.


.gitignore
```diff
@@ -24,3 +24,6 @@ kino-*.tar
 
 # Temporary files, for example, from tests.
 /tmp/
+
+# Node modules.
+lib/assets/*/node_modules/
```
josevalim, line 29: My suggestion is to have an `assets` folder at the root and compile the artefacts to `lib/assets`. Otherwise we can accidentally publish a Hex package with the whole copy of `node_modules` and so on. :D

lib/assets/data_table/esbuild.config.js
```diff
@@ -0,0 +1,17 @@
...
+  loader: { ".js": "jsx" },
+  watch: {
+    onRebuild(error, result) {
+      if (error) console.error('watch build failed:', error)
+      else console.log('watch build succeeded:', result)
+    },
+  },
+}).then(result => {
+  console.log('watching...')
+})
```
jonatanklosko, line 17: Could we also just define an npm script with `esbuild ... --watch`?

lib/kino/table.ex
```diff
@@ -138,19 +142,46 @@ defmodule Kino.Table do
...
+
+  defp type_of(data) do
+    case Float.parse(data) do
+      {_, ""} -> "number"
+      _ -> if date?(data) or date_time?(data), do: "date", else: "text"
+    end
+  end
+
+  defp date?(value), do: match?({:ok, _}, Date.from_iso8601(value))
+  defp date_time?(value), do: match?({:ok, _, _}, DateTime.from_iso8601(value))
```
jonatanklosko, line 183:
```suggestion
  defp type_of(data) do
    cond do
      number?(data) -> "number"
      date?(data) or date_time?(data) -> "date"
      true -> "text"
    end
  end

  defp number?(value), do: match?({_, ""}, Float.parse(value))
  defp date?(value), do: match?({:ok, _}, Date.from_iso8601(value))
  defp date_time?(value), do: match?({:ok, _, _}, DateTime.from_iso8601(value))
```

lib/kino/table.ex
```diff
@@ -138,19 +142,46 @@ defmodule Kino.Table do
 
     ctx = assign(ctx, state: state, key_to_string: key_to_string)
 
+    sample_data = List.first(rows)
```
jonatanklosko, line 145: We also need to handle no rows.

lib/assets/data_table/main.js
```diff
@@ -1,163 +1,310 @@
...
-        ctx.pushEvent("show_page", { page: this.data.content.page - 1 });
-      },
+    );
+  }
 
-      next() {
-        this.loading = true;
-        ctx.pushEvent("show_page", { page: this.data.content.page + 1 });
-      },
+  function App() {
```
jonatanklosko, line 178: Maybe let's move all the components outside `init` and pass the necessary data/ctx as props to the `App` component? That should make things a bit more isolated and easier to analyse :)

lib/assets/data_table/main.js
```diff
@@ -1,163 +1,310 @@
...
+
+    useEffect(() => {
+      const icon = content.order === "asc" ? "arrowUp" : "arrowDown";
+      const newColumns = columns.map((header) => {
+        if (header.id === content.order_by) {
+          return { ...header, overlayIcon: icon };
+        } else {
+          return { ...header, overlayIcon: null };
+        }
+      });
```
jonatanklosko, line 246:
```suggestion
      const newColumns = columns.map((header) => ({
        ...header,
        overlayIcon: header.id === content.order_by ? icon : null
      });
```

lib/assets/data_table/main.js
```diff
@@ -1,163 +1,310 @@
...
+      setColumns(newColumns);
+    }, [content]);
```
jonatanklosko, line 248: We can be more specific to avoid reevaluations, especially that content is an object

```suggestion
    }, [content.order, content.order_by]);
```

assets/data_table/main.js
```diff
@@ -0,0 +1,317 @@
...
+};
+
+export function init(ctx, data) {
+  ctx.importCSS("main.css");
+  ctx.importCSS(
+    "https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap"
+  );
+
+  const container = document.getElementById("root");
+  const root = createRoot(container);
```
jonatanklosko, line 76:
```suggestion
  const root = createRoot(ctx.root);
```

assets/data_table/main.js
```diff
@@ -0,0 +1,317 @@
...
+      )}
+      {!hasData && <p className="no-data">No data</p>}
+      <div id="portal" />
+    </div>
+  );
+}
+
+function Refetch() {
+  const refetch = () => {
+    ctx.pushEvent("refetch");
```
jonatanklosko, line 227: We need to pass `ctx` as a prop.

assets/data_table/main.js
```diff
@@ -0,0 +1,317 @@
...
+      <div className="navigation">
+        <div className="navigation__info">
+          <h2 className="navigation__name">{data.name}</h2>
+          <span className="navigation__details">
+            {totalRows || "?"} entries
+          </span>
+        </div>
+        <div className="navigation__space"></div>
+        {hasRefetch && <Refetch />}
+        <Search toggleSearch={toggleSearch} />
```
jonatanklosko, line 184: Let's name these `RefetchButton` and `SearchButton` for clarity.

assets/data_table/main.js
```diff
@@ -0,0 +1,317 @@
...
+        <RiSearch2Line />
+      </button>
+    </span>
+  );
+}
+
+function Limit({ limit, totalRows, ctx }) {
+  const handleChange = (event) => {
+    ctx.pushEvent("limit", { limit: parseInt(event.target.value) });
+  };
```
jonatanklosko, line 254: I would pass `onChange` to the component and then invoke it with `parseInt(event.target.value)`. This way this component is less concerned with the actual operation.

assets/data_table/main.js
```diff
@@ -0,0 +1,317 @@
...
+}
+
+function Pagination({ page, maxPage, ctx }) {
+  const prev = () => {
+    ctx.pushEvent("show_page", { page: page - 1 });
+  };
+
+  const next = () => {
+    ctx.pushEvent("show_page", { page: page + 1 });
+  };
```
jonatanklosko, line 278: Similarly here we can do `onPrev` and `onNext`.

assets/data_table/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "private": true,
+  "scripts": {
+    "build": "esbuild main.js --bundle --minify --outfile=../../lib/assets/data_table/build/main.js --format=esm --loader:.js=jsx --watch"
```
jonatanklosko, line 4: We can have both `build` and `start` (or `dev`) that watches.

lib/kino/table.ex
```diff
@@ -138,19 +143,48 @@ defmodule Kino.Table do
...
     ctx = assign(ctx, state: state, key_to_string: key_to_string)
 
+    sample_data = if rows, do: List.first(rows)
+
+    columns =
+      if sample_data,
+        do:
+          infer_types(sample_data)
+          |> Enum.zip_with(columns, fn type, column -> Map.put_new(column, :type, type) end),
+        else: columns
```
jonatanklosko, line 153:
```suggestion
    columns =
      if sample_row = List.first(rows) do
        sample_row
        |> infer_types()
        |> Enum.zip_with(columns, fn type, column ->
          Map.put_new(column, :type, type)
        end)
      else
        columns
      end
```

lib/kino/table.ex
```diff
@@ -138,19 +143,48 @@ defmodule Kino.Table do
...
-      order_by: key_to_string[ctx.assigns.order_by]
+      order_by: key_to_string[ctx.assigns.order_by],
+      limit: ctx.assigns.limit
     }
 
     {content, ctx}
   end
 
+  defp infer_types(row) do
+    fields = Map.values(row.fields) |> Enum.map(&String.trim(&1, "\""))
```
jonatanklosko, line 170: Do we still need this, since we no longer inspect?

hugobarauna:
A whole new data table, this is awesome!

I reviewed the PR from the perspective of a user (user experience etc.). I have just one comment:

## Width of columns

Is it possible for the default width of columns to be enough to contain the column headers?

Using the code from your video:

```elixir
keys = ~w(registered_name initial_call reductions stack_size)a

processes = 
  for pid <- Process.list(),
  info = Process.info(pid, keys),
  do: info

Kino.DataTable.new(processes)
```

The default columns look like this:

CleanShot 2022-11-22 at 10 21 38

The problem is that right after plotting the grid, and I'd need to manually resize the columns, where it looks like there's enough space to accommodate all of them by default.

Here's what they look like once I manually resized them:

CleanShot 2022-11-22 at 10 25 15

cristineguadelupe:
> ```elixir
> keys = ~w(registered_name initial_call reductions stack_size)a
> 
> processes = 
>   for pid <- Process.list(),
>   info = Process.info(pid, keys),
>   do: info
> 
> Kino.DataTable.new(processes)
> ```

Hey @hugobarauna! Which browser are you using? This is already the expected behavior. I need to check why it's not working there. 🤔 

https://user-images.githubusercontent.com/5660941/203362783-504aa8d2-c846-4df2-abc0-aeb2b7b0cd10.mp4

hugobarauna:
Regarding:

> Column selection conflicts with sorting: Both are triggered by onHeaderClicked. I don't think this is a big problem, so I kept the column selection active. This can also be solved by manually controlling the selections

If it makes sense, make we could just disable columns selection for now, to avoid that (UX) conflict? I mean, if we're not using that column selection for something yet, disabling could be better.

jonatanklosko (approved):
Awesome 🐈

PR #213, opened 2022-11-26
Header menu - orderBy

cristineguadelupe:
https://user-images.githubusercontent.com/5660941/204110958-a9d2dfb0-d9d3-479f-9979-909adedd8000.mp4


assets/data_table/main.js
```diff
@@ -303,14 +346,18 @@ function Pagination({ page, maxPage, onPrev, onNext }) {
...
+  return (
+    <div className="header-menu" {...layerProps}>
+      <div className="header-menu-item" onClick={() => orderBy("asc")}>
+        ascending
+      </div>
+      <div className="header-menu-item" onClick={() => orderBy("desc")}>
+        descending
+      </div>
+      <div className="header-menu-item" onClick={() => orderBy(null)}>
+        none
```
josevalim, line 359: We will likely add other commands later. Should this say:

```suggestion
        Sort: ascending
      </div>
      <div className="header-menu-item" onClick={() => orderBy("desc")}>
        Sort: descending
      </div>
      <div className="header-menu-item" onClick={() => orderBy(null)}>
        Sort: none
```

Additionally we could use icons, but I think it is probably overkill?

cristineguadelupe: Well, we already have icons on the header and soon we'll have the summaries. I'm not sure we should add more icons... I think it would be too much information

assets/data_table/main.js
```diff
@@ -169,6 +194,21 @@ function App({ ctx, data }) {
...
+      ...header,
+      themeOverride: header.id === currentMenu ? themeOverride : null,
+    }));
+    setColumns(newColumns);
+  }, [showMenu]);
+
+  useEffect(() => {
+    const showMenu = menu ? true : false;
+    setShowMenu(showMenu);
+  }, [menu]);
```
jonatanklosko, line 210: Can't we just determine it on the fly?

```suggestion
  const showMenu = !!menu;
```

cristineguadelupe: Yep, you're right. The problem I'm facing is that clicking on the header triggers the menu, but it also triggers the  `onOutsideClick` immediatly. So when I use a const I get` false -> true -> false` with only one click. The header is out of the menu bounds. I think it's possible what you're suggesting I just couldn't get there.

cristineguadelupe: Ok, Glide has two examples, one using a const and another using only one `useEffect`. The main difference is that in both it's not possible to close the menu from the menu button. Tbh, I prefer our behaviour. Mainly coz in our case you have to click outside the menu but inside the iframe to close it.

https://user-images.githubusercontent.com/5660941/204189636-9274d67f-2336-465d-96db-5b4cc49b8f8e.mov


https://user-images.githubusercontent.com/5660941/204189695-3f941181-9e8b-4e1b-b007-4535c0492c4b.mov

jonatanklosko: > The problem I'm facing is that clicking on the header triggers the menu, but it also triggers the onOutsideClick immediatly

Ah I see, so it makes sense that deferring `showMenu` by one render cycle helps, let's do that for now!

jonatanklosko (approved):

PR #219, opened 2022-12-05
Release v0.8.0

jonatanklosko:
We depend on Nx `~> 0.1`, so we don't need to wait for the next release.


lib/kino.ex
```diff
@@ -112,6 +112,22 @@ defmodule Kino do
...
+  > the following guidelines:
+  >
+  >   * prefix package name with `kino_`, usually followed by the
+  >     name of the integration, such as `kino_vega_lite`, `kino_ecto`
+  >
+  >   * namespace all modules under `KinoExample`, not `Kino.Example`.
+  >     Note that official packages maintained by the Livebook team
+  >     expose public APIs under `Kino.`, because they are essentially
+  >     direct extensions of `Kino` and we make sure no conflicting
+  >     modules exist. Unofficial packages should follow the regular
```
jonatanklosko, line 128:
```suggestion
  >     modules exist. Unofficial packages should follow the usual
```

lib/kino.ex
```diff
@@ -112,6 +112,22 @@ defmodule Kino do
...
+  >
+  >   * prefix package name with `kino_`, usually followed by the
+  >     name of the integration, such as `kino_vega_lite`, `kino_ecto`
+  >
+  >   * namespace all modules under `KinoExample`, not `Kino.Example`.
+  >     Note that official packages maintained by the Livebook team
+  >     expose public APIs under `Kino.`, because they are essentially
+  >     direct extensions of `Kino` and we make sure no conflicting
+  >     modules exist. Unofficial packages should follow the usual
+  >     Elixir conventions with respect to modules names
```
jonatanklosko, line 129:
```suggestion
  >     Elixir conventions with respect to module names
```

josevalim (approved):
And I think we can ship nx_image too!

PR #220, opened 2022-12-05
Add `monospace` option to `Kino.Input.textarea`

zolrath:
Allow `Kino.Input.textarea` inputs to be given a new option, `monospace`.
When `monospace` is set to `true`, this allows Livebook to render the textarea with a monospaced font.

This is to support https://github.com/livebook-dev/livebook/pull/1565

When combined with the pull request above it results in:

image

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -71,11 +71,19 @@ defmodule Kino.Input do
   ## Options
 
     * `:default` - the initial input value. Defaults to `""`
+    * `:monospace` - display the textarea's contents in a monospace font. Defaults to `false`
   """
   @spec textarea(String.t(), keyword()) :: t()
   def textarea(label, opts \\ []) when is_binary(label) and is_list(opts) do
     default = opts |> Keyword.get(:default, "") |> to_string()
-    new(%{type: :textarea, label: label, default: default})
+    monospace = opts |> Keyword.get(:monospace, false)
+
+    new(%{
+      type: :textarea,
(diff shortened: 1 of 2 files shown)


lib/kino/input.ex
```diff
@@ -71,11 +71,19 @@ defmodule Kino.Input do
   ## Options
 
     * `:default` - the initial input value. Defaults to `""`
+    * `:monospace` - display the textarea's contents in a monospace font. Defaults to `false`
```
jonatanklosko, line 74:
```suggestion
    * `:default` - the initial input value. Defaults to `""`

    * `:monospace` - whether to use a monospace font inside the textarea.
      Defaults to `false`
```

lib/kino/input.ex
```diff
@@ -71,11 +71,19 @@ defmodule Kino.Input do
...
   """
   @spec textarea(String.t(), keyword()) :: t()
   def textarea(label, opts \\ []) when is_binary(label) and is_list(opts) do
     default = opts |> Keyword.get(:default, "") |> to_string()
-    new(%{type: :textarea, label: label, default: default})
+    monospace = opts |> Keyword.get(:monospace, false)
```
jonatanklosko, line 79:
```suggestion
    monospace = Keyword.get(opts, :monospace, false)
```

(we usually use pipe if there are more than a single function call :))

jonatanklosko (approved):
Thanks!

PR #221, opened 2022-12-06
Summary header

ghost:
This PR aims two things!

- Draw a custom header with the summary data when available

- A naive integration with Explorer so we can get an overview of the data we'll need and decide how to handle it

CleanShot 2022-12-06 at 00 26 30@2x

https://user-images.githubusercontent.com/5660941/205808368-bbe03dee-bb29-4ecb-b0a1-b38ef1df25b6.mp4

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -240,6 +240,6 @@ defmodule Kino.DataTable do
   defp value_to_string(value) when is_atom(value), do: inspect(value)
 
   defp value_to_string(value) do
-    if String.Chars.impl_for(value), do: "#{value}", else: "#{inspect(value)}"
+    if mod = String.Chars.impl_for(value), do: mod.to_string(value), else: inspect(value)
   end
 end


lib/kino/table.ex
```diff
@@ -83,7 +83,8 @@ defmodule Kino.Table do
     payload = %{
       name: ctx.assigns.info.name,
       features: ctx.assigns.info.features,
-      content: ctx.assigns.content
+      content: ctx.assigns.content,
+      summary: ctx.assigns.info.summary
```
jonatanklosko, line 87: We should also added it to the typespec :)

mix.exs
```diff
@@ -33,6 +33,7 @@ defmodule Kino.MixProject do
     [
       {:table, "~> 0.1.2"},
       {:nx, "~> 0.1", optional: true},
+      {:explorer, "~> 0.4.0", optional: true},
```
jonatanklosko, line 36: This should point to main for now, because we don't have `DF.describe` in 0.4.0, and this way we don't forget to bump to `0.5.0` once it's out.

lib/kino/data_table.ex
```diff
@@ -240,4 +241,54 @@ defmodule Kino.DataTable do
   defp value_to_string(value) do
     if String.Chars.impl_for(value), do: "#{value}", else: "#{inspect(value)}"
   end
+
+  defp summary(%Explorer.DataFrame{} = df) do
+    describe =
+      df
+      |> Explorer.DataFrame.describe()
+      |> Explorer.DataFrame.slice([1, 3, 7])
```
jonatanklosko, line 249: Can we switch it such that we use statistic names rather than indices, such that we don't rely on the order?

assets/data_table/main.js
```diff
@@ -106,14 +109,157 @@ function App({ ctx, data }) {
...
+        Mean: summary.mean?.toFixed(2) ?? "",
+        Nulls: summary.nulls ?? "",
+      };
+
+      const categoricalInfo = {
+        Unique: summary.unique ?? "",
+        Top: summary.top ?? "",
+        "Top freq": summary.top_freq ?? "",
+        Nulls: summary.nulls ?? "",
+      };
```
josevalim, line 224: This looks great but I think we can move this to the server, what do you think?

This way, the server would send something like this:

```
[
  %{key: "Unique", value: "321"},
  ...
]
```

And all we do is to render the given keys and values? This means if we want to change how what is rendered, we just need to change the Elixir side and we don't have to write more JavaScript. :)

mix.exs
josevalim: The dependency on Explorer is temporary... we should talk how we want to handle this via Table/Explorer. :) There are three functionalities:

1. Aggregations
2. Summaries
3. Sorting

We can have a convo later today!

lib/kino/data_table.ex
```diff
@@ -237,6 +237,8 @@ defmodule Kino.DataTable do
     end
   end
 
+  defp value_to_string(value) when is_atom(value), do: inspect(value)
+
   defp value_to_string(value) do
     if String.Chars.impl_for(value), do: "#{value}", else: "#{inspect(value)}"
```
josevalim, line 243:
```suggestion
    if mod = String.Chars.impl_for(value), do: mod.to_string(value), else: inspect(value)
```

Issue #217, opened 2022-12-01
Textarea content is too long, is it appropriate to truncate?


watsy0007:
<img width="774" alt="image" src="https://user-images.githubusercontent.com/822237/205010732-b8d7d8ce-e746-4d98-b9df-d5a569dd31a6.png">

is it appropriate to truncate in preview mode, show all content when clicked ?
or add switch in settings to display all content adaptively.

PR #223, opened 2022-12-08
Data table improvements and fixes

cristineguadelupe:
- `allowWrapping` is now `false`, as we don't have dynamic heights, the problems don't outweigh the benefits
- We now have a `maxColumnAutoWidth` which fixes the #217 
- `getCellsForSelection` misscalculates the offset depending on how the rows are selected. I've temporarily fixed that on our side. Now regardless of how the selection is made, the cells are correctly copied.
- I've removed the checkboxes from `rowMarkers`. They are now clickable numbers and have a hover effect. IMO this makes the table look more minimalist and integrated with the whole Livebook

https://user-images.githubusercontent.com/5660941/206335107-9bd74fe2-f3c8-4fec-af94-04386f89eaf0.mp4


assets/data_table/main.js
```diff
@@ -322,6 +327,46 @@ function App({ ctx, data }) {
     []
   );
 
+  const onItemHovered = useCallback((args) => {
+    const [col, row] = args.location;
+    row === -1 && col === -1 && args.kind === "header"
+      ? setHoverRows([...Array(rows).keys()])
```
jonatanklosko, line 333: You can use this

```javascript
function times(n, fn) {
  return Array.from({ length: n }, (_, index) => fn(index));
}
```

assets/data_table/main.js
```diff
@@ -322,6 +327,46 @@ function App({ ctx, data }) {
...
+      : col === -1 && args.kind === "cell"
+      ? setHoverRows([row])
+      : setHoverRows(null);
```
jonatanklosko, line 336: Please make this an if-ifelse-else :)

josevalim (approved):

PR #224, opened 2022-12-12
Kino Explorer - new api and data summary

cristineguadelupe:
This PR updates #42 to the new Explorer api and adds the summary.


josevalim:
I talked to @cristineguadelupe and we decided to keep `Kino.Explorer` in `Kino` for now, to ease development, BUT we should extract it out before a new Kino version is out.

lib/kino/table.ex
```diff
@@ -1,6 +1,15 @@
...
   @moduledoc false
 
+  @type numeric_summary :: %{min: float(), max: float(), mean: float(), nulls: float()}
+
+  @type categorical_summary :: %{
+          top: String.t(),
+          top_freq: integer(),
+          unique: integer(),
+          nulls: float()
+        }
```
jonatanklosko, line 11: We can format the summary value into string on the server right? Then we can generalize to (which I believe we already handle on the client):

```elixir
@type summary :: %{String.t() => String.t()}
```

lib/kino.ex
```diff
@@ -24,6 +24,13 @@ defmodule Kino do
 
       Kino.DataTable.new(data)
 
+  ### Kino.Explorer
+
+   `Kino.Explorer` implements a data table output for [Explorer](https://github.com/elixir-nx/explorer)
+   data frames:
+       df = Explorer.Datasets.fossil_fuels()
+       Kino.Explorer.new(df)
+
```
jonatanklosko, line 33: Let's not add, so we don't need to remove later :)

```suggestion
```

lib/kino/explorer.ex
```diff
@@ -0,0 +1,129 @@
...
+  ## Examples
+      df = Explorer.Datasets.fossil_fuels()
+      Kino.Explorer.new(df)
+  """
+
+  @behaviour Kino.Table
+
+  @type t :: Kino.JS.Live.t()
+
+  @compile {:no_warn_undefined, Explorer.DataFrame}
```
jonatanklosko, line 13:
```suggestion
  @compile {:no_warn_undefined, Explorer.DataFrame}
  @compile {:no_warn_undefined, Explorer.Series}
```

lib/kino/explorer.ex
```diff
@@ -0,0 +1,129 @@
...
+        %{"counts" => [top_freq], "values" => [top]} = most_frequent(series)
+        {label, %{nulls: nulls, top: top, top_freq: top_freq, unique: count_unique(series)}}
+      end
+    end
+  end
+
+  defp describe(data) do
+    data
+    |> Explorer.DataFrame.describe()
+    |> Explorer.DataFrame.slice([1, 3, 7])
```
jonatanklosko, line 102: Is there are way we can rely on names and not indices?

cristineguadelupe: Not directly. To make it clear we can have something like `mean_idx = 1, min_idx = 3...` or skip the slice and go with `for {column, [_count, mean, _std, min, ... max]} <- describe,`

lib/kino/table.ex
cristineguadelupe: I don't get it... 🤔 Do you mean a single string with all the values?

jonatanklosko: Nono, that's just the map typespec. So it would look something like `%{"min" => "1.234", "max" => "4.144", ...}`.

cristineguadelupe: Oh! Got it. Yep! We can have strings 🙈

cristineguadelupe: Why not `optional(:summary) => %{String.t() => String.t()}` directly?

jonatanklosko: That works, sure :)

PR #229, opened 2022-12-20
Data Table - Filtering support for Kino.Explorer

cristineguadelupe:
Working on final touch-ups

https://user-images.githubusercontent.com/5660941/208582993-1f4f8e6a-e49d-45f8-b780-409165e70212.mp4


josevalim:
To reduce the amount of buttons on the interface, what do you think about "Filters:" starting with a value called "None", similar to Sort? If you want to cancel the filter, you can revert it to "None". Initially it will look like this:

image

Notice I aligned the select to the right  (I think we should align all of them to the right).

Then, once you click a filter, we will show this and input with the button inside the input:

image

But probably on the right side?

josevalim:
Another nitpick: I also think we are using the Inter font too much here. In this dropdown, I would perhaps use the regular font (like we do in Kino.VegaLite dropdown). I would perhaps also use regular font on the title, subtitle, "Next", "Prev", and  "Show".

cristineguadelupe:
Ok! I agree with both your suggestions.

assets/data_table/main.js
```diff
@@ -64,10 +64,23 @@ const theme = {
...
 
+const filtering = {
+  numeric: [
+    "less",
+    "less_equal",
+    "equal",
+    "not_equal",
+    "greater_equal",
+    "greater",
+  ],
```
josevalim, line 75: In the select, let's show the labels without underscore. :)

assets/data_table/main.js
```diff
@@ -422,8 +464,18 @@ function App({ ctx, data }) {
         <div className="navigation__info">
           <h2 className="navigation__name">{data.name}</h2>
           <span className="navigation__details">
-            {totalRows || "?"} entries
+            {totalRows || "?"} {totalRows > 1 ? "entries" : "entry"}
```
jonatanklosko, line 467: We also need to handle "0 entries"

```suggestion
            {totalRows || "?"} {totalRows === 1 ? "entry" : "entries"}
```

assets/data_table/main.js
```diff
@@ -271,10 +301,20 @@ function App({ ctx, data }) {
...
     ctx.pushEvent("order_by", { key, order: order ?? "asc" });
   };
 
+  const filterBy = (filter, value) => {
+    const { id, title } = menu
+      ? columns[menu.column]
+      : { id: null, title: null };
+    const key = filter ? id : null;
+    const filterMessage = filter ? `(${title} ${filter} ${value})` : null;
+    setFilter({ filter: "none", filterValue: "", filterMessage });
```
jonatanklosko, line 314: Why are we setting "none" if there is `filter`?

lib/kino/table.ex
```diff
@@ -3,14 +3,25 @@ defmodule Kino.Table do
...
   @type rows_spec :: %{
           offset: non_neg_integer(),
           limit: pos_integer(),
           order_by: nil | term(),
-          order: :asc | :desc
+          order: :asc | :desc,
+          filter_by: nil | term(),
+          filter_value: nil | term(),
+          filter:
+            :none
```
jonatanklosko, line 17: I think we should use `nil` for optional entries, this way if we add add something else to the rows_spec later it's enough to check `spec[:option] == nil` to cover both the none case and the option missing altogether (backwards compatibility).

Same for `order_by` :)

lib/kino/table.ex
```diff
@@ -124,6 +138,28 @@ defmodule Kino.Table do
...
+  def handle_event("filter_by", %{"key" => key_string, "filter" => filter, "value" => value}, ctx) do
+    filter = String.to_existing_atom(filter)
+
+    ctx =
+      if key_string do
+        # Lookup key by the string representation received from the client
+        ctx.assigns.key_to_string
+        |> Enum.find(&match?({_key, ^key_string}, &1))
+        |> case do
+          {key, _key_string} ->
```
jonatanklosko, line 150: Let's move this to a `lookup_key` function and reuse for order :)

assets/data_table/main.js
```diff
@@ -565,20 +630,131 @@ function Pagination({ page, maxPage, onPrev, onNext }) {
...
+
+  function castFilterValue(value) {
+    if (isNumber) {
+      return parseFloat(value);
+    }
+    if (isDate) {
+      return new Date(value).toISOString();
+    } else {
+      return value;
+    }
```
jonatanklosko, line 695:
```suggestion
    }

    return value;
```

assets/data_table/main.js
```diff
@@ -64,10 +64,23 @@ const theme = {
...
+  categorical: ["equal", "contains", "not_equal"],
+};
+
 export function init(ctx, data) {
   ctx.importCSS("main.css");
   ctx.importCSS(
-    "https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap"
+    "https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap",
+    "https://fonts.googleapis.com/css2?family=Inter:wght@400;500&display=swap"
   );
```
jonatanklosko, line 84:
```suggestion
    "https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600&display=swap"
  );
  ctx.importCSS(
    "https://fonts.googleapis.com/css2?family=Inter:wght@400;500&display=swap"
  );
```

jonatanklosko:
Also, sorting is reflected unless I open the select a second time:

https://user-images.githubusercontent.com/17034772/208931752-8a52af46-cc20-4a1b-beac-16691c2839b7.mov

cristineguadelupe:
Suggestions + fixes + cumulative filtering + prefill for filter and sort


https://user-images.githubusercontent.com/5660941/209412987-a29ac056-972a-4c73-8503-aeb2355d1935.mp4

josevalim:
One last nitpick: can you make it so the `v` of the select and the input touch the red line:

Screenshot 2022-12-23 at 22 13 49

Then we can ship it :)

josevalim (approved):

josevalim (approved):

lib/kino/table.ex
```diff
@@ -3,14 +3,15 @@ defmodule Kino.Table do
...
+          features: list(:refetch | :pagination | :sorting | :filtering)
         }
 
   @type rows_spec :: %{
           offset: non_neg_integer(),
           limit: pos_integer(),
           order_by: nil | term(),
-          order: :asc | :desc
+          order: :asc | :desc,
+          filters: nil | list(%{String.t() => String.t()})
```
jonatanklosko, line 14: `Kino.Table` has full control over the filter, so we should use a more specific type than `%{String.t() => String.t()}`. Something like `%{type : :equal | :less | ..., value: term(), column: term()}`.

Also, given it is a list now, we don't really need `nil`.

Sidenote: perhaps we should change order to `order: nil | %{direction: :asc | :desc, column: term()}` for consistency, wdyt? At this point it's fine to do it, since `Kino.Table` is not a public API yet :)

cristineguadelupe: Nice! Gonna do this in #237 coz I'll already be touching the front

PR #236, opened 2022-12-25
Updates filters and order

cristineguadelupe:
When changing a column's filter, adds it to the end as a new filter instead of updating it in place


lib/kino/table.ex
```diff
@@ -120,20 +120,16 @@ defmodule Kino.Table do
 
   def handle_event("filter_by", %{"key" => key} = filter, ctx) do
     filters = if ctx.assigns.filters, do: ctx.assigns.filters, else: []
-    idx = Enum.find_index(filters, &(&1["key"] == key))
-    updated_filters = if idx, do: List.replace_at(filters, idx, filter), else: filters ++ [filter]
-    keys = Enum.map(filters, &lookup_key(ctx, &1["key"])) |> Enum.all?()
+    updated_filters = Enum.reject(filters, &(&1["key"] == key)) |> Kernel.++([filter])
+    keys = Enum.map(updated_filters, &lookup_key(ctx, &1["key"])) |> Enum.all?()
```
jonatanklosko, line 124: We only need to check the key received in this event, the other ones should be fine.

FWIW I can't really think of a scenario where we would receive an invalid stringified key, but I see that's what I did for order_by initially, probably just in case 🙈

cristineguadelupe: Yap and talking about lookup, in the filters I end up using it only for validation because I already send the column title. But we could follow the same approach as `order_by ` passing only the key and mapping…

jonatanklosko: We need the original key when calling `get_data` on the `Kino.Table` implementation, no?

cristineguadelupe: Yep, but just like in `order`, we could have a map to -> map back.
So, I think the only advantage is consistency if you think it’s worth it.

jonatanklosko: So what you mean is that we don't use `key_to_string` for filters? We want to always map columns to strings when sending them to the client, because columns can technically be any term.

cristineguadelupe: Exactly! As in Explorer we have `name` as key and `to_string(name)` as title and `filter_with` already uses the string version, I’m ignoring it. But we’ll make the api public soon! I’m just sending both and using key for comparisons and title for the filter itself. I guess I should change it 🙈

jonatanklosko: `Kino.Table` is Explorer-agnostic, so we should not make any assumptions :)

cristineguadelupe: Yep!

cristineguadelupe:
News filters at the last + https://github.com/livebook-dev/kino/pull/229#discussion_r1056932621 + #237

assets/data_table/main.js
```diff
@@ -411,7 +411,7 @@ function App({ ctx, data }) {
         setCurrentMenuForm(currentMenuForm);
       }
     },
-    [filters, content.order_by, content.order]
+    [filters, content.order]
```
jonatanklosko, line 414: We send the whole `content` object on every update, so both `content.filters` and `content.order` are different objects (that is, different object references), similarly to how `{} === {}` is actually `false`. `useCallback` and friends only do flat comparison of the dependencies we specify, so the callback will be invalidated whenever we set new `content`, even if the object contents are the same.

So in most cases we don't want to pass objects as dependencies to `use*`, instead specify the individual properties that we care about. For something like filters it's tricky because the number of items may change. One option is to use `JSON.stringify(filters)`, this way we compare the content exactly, but it is a little bit more costly, because we stringify on every render and do string comparison. There's also the option to use `useRef`, such that the callback uses a mutable state.

All that said, does this callback actually depend on filters/order?

cristineguadelupe: Hey! In this case I'm using the deps to force the `getCurrentMenuForm` to be reevaluated. What could I do instead? 🤔

assets/data_table/main.js
```diff
@@ -479,13 +454,13 @@ function App({ ctx, data }) {
...
-    const icon = content.order === "asc" ? "arrowUp" : "arrowDown";
+    const icon = content.order?.direction === "asc" ? "arrowUp" : "arrowDown";
     const newColumns = columns.map((header) => ({
       ...header,
-      overlayIcon: header.id === content.order_by ? icon : null,
+      overlayIcon: header.id === content.order?.key ? icon : null,
     }));
     setColumns(newColumns);
-  }, [content.order, content.order_by]);
+  }, [content.order]);
```
jonatanklosko, line 463:
```suggestion
  }, [content.order?.direction, content.order?.key]);
```

lib/kino/table.ex
```diff
@@ -108,36 +112,36 @@ defmodule Kino.Table do
...
-    filters = if ctx.assigns.filters, do: ctx.assigns.filters, else: []
-    idx = Enum.find_index(filters, &(&1["key"] == key))
-    updated_filters = if idx, do: List.replace_at(filters, idx, filter), else: filters ++ [filter]
-    keys = Enum.map(filters, &lookup_key(ctx, &1["key"])) |> Enum.all?()
-    ctx = if keys, do: assign(ctx, filters: updated_filters), else: ctx
+  def handle_event("filter_by", %{"key" => key_string, "filter" => filter, "value" => value}, ctx) do
+    updated_filters =
+      Enum.reject(ctx.assigns.content.filters, &(&1.key == key_string))
+      |> Kernel.++([%{key: key_string, filter: filter, value: value}])
+      |> Enum.map(&%{&1 | key: lookup_key(ctx, &1.key)})
```
jonatanklosko, line 129: We can work on `ctx.assigns.filters` rather than `ctx.assigns.content.filters`, and then we don't need this additional `Enum.map`, right?

lib/kino/table.ex
```diff
@@ -108,36 +112,36 @@ defmodule Kino.Table do
...
+
+    ctx = if lookup_key(ctx, key_string), do: assign(ctx, filters: updated_filters), else: ctx
```
jonatanklosko, line 131: Let's move `updated_filters =` inside the if, then we only need one `lookup_key` in total.

lib/kino/table.ex
cristineguadelupe: Yep!

assets/data_table/main.js
```diff
@@ -585,32 +560,42 @@ function App({ ctx, data }) {
...
           <button className="button" onClick={resetFilters}>
             <span>Reset filters</span>
           </button>
+          <button
+            className="button"
+            onClick={() =>
+              filterBy({ ...lastFilter, filter: null, value: null })
+            }
+          >
+            <span>Undo last filter</span>
```
jonatanklosko, line 586: If there are 0 rows in the dataframe, would we still show this? If so, we need to make sure we actually have `lastFilter` :)

jonatanklosko (approved):
Just one more comment :)

assets/data_table/main.js
cristineguadelupe: I'm not sure I understand your question. We only show this if a filter or a combination of filters results in 0 rows. If the original df has 0 rows, we show "no data" instead.

jonatanklosko: We show "no data" if there are no columns, but it is possible that we have 0 rows and still show columns. Perhaps not with explorer, but I think this could be the case for database queries that return no rows.

cristineguadelupe: Gotcha! I'll ensure there are filters!

PR #240, opened 2022-12-29
Columnar data

cristineguadelupe:
ATM Explorer as columnar and DataTable as row based. WDYT? Let’s support both?


cristineguadelupe:
And if we go with both, do we keep the DataTable as row-based or change it to columnar?

lib/kino/table.ex
```diff
@@ -45,7 +49,7 @@ defmodule Kino.Table do
               {:ok,
                %{
                  columns: list(column()),
-                 rows: list(row()),
+                 data: columnar() | row_based(),
```
jonatanklosko, line 52: We decided we don't want to send rows as field maps, because we repeat fields for every row. So in either case it's a list of lists, to distinguish they need to give us `{:rows, [[...], ...]}` or `{:columns, [[...], ...]}` (and we assume it matches the order of columns).

lib/kino/table.ex
```diff
@@ -31,6 +31,10 @@ defmodule Kino.Table do
           fields: list(%{term() => String.t()})
         }
 
+  @type row_based :: list(row())
+
+  @type columnar :: %{term() => list(String.t())}
```
josevalim, line 36: Why is it `String.t()`? Isn't it anything encodable to json?

jonatanklosko: We encode all values to strings on Elixir side already, this matches `fields:` above :)

lib/kino/table.ex
```diff
@@ -49,7 +40,7 @@ defmodule Kino.Table do
               {:ok,
                %{
                  columns: list(column()),
-                 data: columnar() | row_based(),
+                 data: {:columnar | :row_based, list(list(String.t()))},
```
jonatanklosko, line 43: Let's match the wording in `Table.Reader`

```suggestion
                 data: {:columns | :rows, list(list(String.t()))},
```

lib/kino/table.ex
```diff
@@ -162,19 +153,16 @@ defmodule Kino.Table do
...
 
-    row_based = is_list(data)
-    page_length = if row_based, do: length(data), else: Enum.at(data, 0) |> elem(1) |> length()
-
-    sample_data =
-      if row_based, do: List.first(data), else: Map.values(data) |> Enum.map(&List.first/1)
-
+    row_based = orientation == :row_based
+    page_length = if row_based, do: length(data), else: hd(data) |> length()
+    sample_data = if row_based, do: List.first(data), else: Enum.map(data, &List.first(&1))
```
jonatanklosko, line 165: Probably cleaner to do a `case` and return a tuple with `{page_length, sample_data}` :)

assets/data_table/main.js
```diff
@@ -304,9 +304,10 @@ function App({ ctx, data }) {
   const getCellContent = useCallback(
     ([col, row]) => {
       const kind = cellKind[content.columns[col].type] || GridCellKind.Text;
-      const cellData = Array.isArray(content.data)
-        ? content.data[row]?.fields[col]
-        : content.data[col][row];
+      const columnar = content.data_orientation === "columnar";
+      const cellData = columnar
+        ? content.data[col][row]
+        : content.data[row][col];
```
jonatanklosko, line 310: Beautiful :)

cristineguadelupe: Yep :)

jonatanklosko (approved):

PR #242, opened 2023-01-07
Add Kino.JS.Live.monitor/1

mcrumm:
Hello, Livebook devs! This is a very pass-thru function but given the opaque Live type I thought it might be acceptable :)

My use case is a [telemetry handler](https://github.com/mcrumm/kino_telemetry/commit/b981f90cda1534acc365357a501727d76ec387e4) that monitors a VegaLite chart and shuts down when the cell is re-evaluated.


jonatanklosko (approved):
Sounds good, thanks!

PR #248, opened 2023-01-23
Expanded header menu

cristineguadelupe:
Screen Shot 2023-01-23 at 18 51 12


josevalim:
Looking good! Let's increase the vertical gap between elements and add icons:

```
asc: <i class="ri-sort-asc"></i>
desc: <i class="ri-sort-desc"></i>
none: <i class="ri-align-justify"></i>
```

josevalim (approved):

PR #252, opened 2023-01-31
Support non-charlist List values in DataTable

03juan:
My SQL query smartcell crashed with the `List.to_string/1` `ArgumentError` exception when returning an `array_agg` list of timestamps from a postgres query.

This rescues the exception and applies the `value_to_string/1` to each list item before running the whole list through `inspect`.

The test checks for the standard charlist behaviour along with some non-standard values.

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -242,6 +242,15 @@ defmodule Kino.DataTable do
 
   defp value_to_string(value) when is_atom(value), do: inspect(value)
 
+  defp value_to_string(value) when is_list(value) do
+    try do
+      List.to_string(value)
+    rescue
+      ArgumentError ->
+        inspect(value)
+    end
+  end
+
   defp value_to_string(value) do
     if mod = String.Chars.impl_for(value), do: mod.to_string(value), else: inspect(value)
   end
--- a/test/kino/data_table_test.exs
+++ b/test/kino/data_table_test.exs
@@ -114,7 +114,7 @@ defmodule Kino.DataTableTest do
                data: [
                  [
                    <<1, 98>>,
-                   "[~N[2000-01-01 00:00:00], %Kino.DataTableTest.User{__meta__: nil, id: 1, name: \"User\"}]"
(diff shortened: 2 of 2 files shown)


lib/kino/data_table.ex
```diff
@@ -242,6 +242,15 @@ defmodule Kino.DataTable do
 
   defp value_to_string(value) when is_atom(value), do: inspect(value)
 
+  defp value_to_string(value) when is_list(value) do
+    try do
+      List.to_string(value)
+    rescue
+      ArgumentError ->
+        Enum.map(value, &value_to_string/1) |> inspect()
```
josevalim, line 250: Do we need value_to_string if we are going to inspect it anyway?

jonatanklosko: +1, if we inspect we should be consistent with a regular inspect :)

03juan: Not at all, my tired brain thought so... 🙄

josevalim:
@jonatanklosko @cristineguadelupe any concerns here?

josevalim (approved):

jonatanklosko:
Nope, definitely a good call!

03juan:
Fixing test quickly...

test/kino/data_table_test.exs
```diff
@@ -98,6 +98,29 @@ defmodule Kino.DataTableTest do
...
+
+    kino = Kino.DataTable.new(entries)
+    data = connect(kino)
+
+    assert %{
+             content: %{
+               data: [
+                 [
+                   <<1, 98>>,
+                   "[~N[2000-01-01 00:00:00], %Kino.DataTableTest.User{__meta__: nil, id: 1, name: \"User\"}]"
```
jonatanklosko, line 117:
```suggestion
                   ~s/[~N[2000-01-01 00:00:00], %Kino.DataTableTest.User{__meta__: nil, id: 1, name: "User"}]/
```

jonatanklosko (approved):

test/kino/data_table_test.exs
03juan: Thank you

PR #253, opened 2023-02-01
Change file access API

jonatanklosko:
The file input is now used as:

```elixir
value = Kino.Input.read(input)
path = Kino.Input.file_path(value.file_id)

File.read!(path)
```

See [TBA](https://github.com/livebook-dev/livebook/pull/1673) for more details.


josevalim:
We should update the docs for `file` to mention `file_path` has to be called.

josevalim:
Sorry, I meant to say we should include a more complete example. Also one for streaming.

josevalim (approved):

PR #261, opened 2023-03-05
Do automatic stream conversion on listen

josevalim:


josevalim:
The main goal here is to reduce the learning curve. Telling folks they need to convert to a stream before listening is a bit too much, IMO.

josevalim:
@jonatanklosko I understand your concerns now. WDYT about implementing enumerable for Kino.Input and Kino.Control and Kino.Control.stream + tagged_stream are basically mechanisms to merge and tag them? That keeps the API as is (which you want) and makes it more concise (which I want :D).

jonatanklosko:
@josevalim Enumerable works for me!

josevalim:
New PR incoming.

PR #268, opened 2023-03-17
Add concurrent listen

jonatanklosko:
Adds a version of `listen/2` that processes all events concurrently in separate processes.

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -331,6 +331,9 @@ defmodule Kino do
   @doc ~S"""
   Asynchronously consumes a stream with `fun`.
 
+  Note that events are processed by `fun` sequentially. If you want
+  to process them concurrently, use `listen_all/2`.
+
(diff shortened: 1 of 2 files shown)


lib/kino.ex
```diff
@@ -373,6 +376,44 @@ defmodule Kino do
...
+        {:ok, task_supervisor} = start_child(Task.Supervisor)
+        task_supervisor
+      end
+
+      reducer = fn event, task_supervisor ->
+        {[{event, task_supervisor}], task_supervisor}
+      end
+
+      after_fun = fn task_supervisor ->
+        DynamicSupervisor.terminate_child(Kino.DynamicSupervisor, task_supervisor)
```
jonatanklosko, line 404: @josevalim should we have public `Kino.terminate_child`?

josevalim (approved):

PR #269, opened 2023-03-17
Only terminate evaluator on listen exit with non-normal reasons

josevalim:


lib/kino/terminator.ex
```diff
@@ -11,6 +11,45 @@ defmodule Kino.Terminator do
...
+    after
+      Process.group_leader(self(), initial_gl)
+    end
+  end
+
+  @doc """
+  Starts a task that will terminate the parent in case of crashes.
+  """
+  def start_task(parent, fun) do
+    Task.start_link(fn ->
```
jonatanklosko, line 47: We don't want to link, or am I missing it?

josevalim: This is being called inside the supervisor. We still want to link the task to its supervisor. :)

jonatanklosko: Ahh sure it's for the child spec :D

PR #272, opened 2023-03-20
Support for transient frame updates

jonatanklosko:
https://user-images.githubusercontent.com/17034772/226475017-301135f0-19eb-4df5-8d24-29c1809c9567.mov

### DIFF
--- a/lib/kino/frame.ex
+++ b/lib/kino/frame.ex
@@ -171,41 +171,20 @@ defmodule Kino.Frame do
   def handle_cast({:render, term, destination}, state) do
     output = Kino.Render.to_livebook(term)
     put_update(destination, state.ref, [output], :replace)
-
-    state =
-      if destination == :default do
-        %{state | outputs: [output]}
-      else
-        state
-      end
-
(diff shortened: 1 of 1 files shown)


lib/kino/frame.ex
```diff
@@ -64,11 +64,34 @@ defmodule Kino.Frame do
       option is useful when updating frame in response to client
       events, such as form submission
 
+    * `:history` - if the update will be part of the frame history.
```
jonatanklosko, line 67: I'm leaning towards `:transient`, wdyt?

aleDsz: +1 for `:transient`

josevalim: I like `:history` for simplicity. If folks dislike it, `:temporary` perhaps? `:transient` is too fancy. :D

josevalim (approved):

lib/kino/frame.ex
jonatanklosko: I just think `:history` doesn't read as well, especially with `clear`. `:temporary` also works for me :)

aleDsz (approved):

Issue #273, opened 2023-03-28
`handle_call` from the `Kino.JS.Live` behaviour cannot return `:noreply`


Janix4000:
My Kino module uses:
```
use Kino.JS, assets_path: "lib/assets/player"
use Kino.JS.Live
```
and "implements" `handle_call` function:
```
  @impl true
  def handle_call({:create, framerate}, from, ctx) do
    payload = %{framerate: framerate, from: from}

    broadcast_event(ctx, "create", payload)

    {:noreply, ctx}
  end
```
and replies in other callback:
```
  def handle_event("jmuxer_ready", from, ctx) do
    GenServer.reply(from, {:ok, :player_created})
    {:noreply, assign(ctx, jmuxer_ready: true)}
  end
```

However, I cannot return `:noreply` from the `handle_call` callback despite [the documentation](https://hexdocs.pm/kino/Kino.JS.Live.html#c:handle_call/3):
```
[error] GenServer #PID<0.482.0> terminating
** (MatchError) no match of right hand side value: {:noreply, %Kino.JS.Live.Context{assigns: %{clients: ["suokrbya"], jmuxer_ready: false, type: :both}, origin: nil, __private__: %{client_ids_with_pid: %{"kea4nx2qfe3m2xeiaoxoagt5iwz5ahru" => #PID<11798.1126.0>}, monitors_with_client_id: %{#Reference<0.588131328.1499463684.243371> => "kea4nx2qfe3m2xeiaoxoagt5iwz5ahru"}, ref: "7zwxud775xferh3buuko4zchfjqzlrbd"}}}
    (kino 0.8.1) lib/kino/js/live/server.ex:50: Kino.JS.Live.Server.handle_call/3
    (stdlib 4.2) gen_server.erl:1149: :gen_server.try_handle_call/4
    (stdlib 4.2) gen_server.erl:1178: :gen_server.handle_msg/6
    (stdlib 4.2) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Last message (from #PID<0.515.0>): {:create, 25.0}
```
 I think it is due to [these lines](https://github.com/livebook-dev/kino/blob/6a507a3841f3e9bb803eedeebb15031bd590e1e6/lib/kino/js/live/server.ex#L49). 

Is this considered a bug?

jonatanklosko:
Closed in #274.

PR #274, opened 2023-03-28
Support deferred reply in `Kino.JS.Live.handle_call/3`

Janix4000:
Fix the wrong `JS.Live.handle_call/3` behavior, described in #273.

### DIFF
--- a/test/kino/js/live_test.exs
+++ b/test/kino/js/live_test.exs
@@ -25,6 +25,13 @@ defmodule Kino.JS.LiveTest do
       assert count == 1
     end
 
+    test "handle_call/3 :noreply" do
+      kino = LiveCounter.new(0)
+      LiveCounter.bump(kino, 1)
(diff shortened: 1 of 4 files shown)


test/kino/js/live_test.exs
```diff
@@ -25,6 +25,13 @@ defmodule Kino.JS.LiveTest do
       assert count == 1
     end
 
+    test "handle_call/3 :noreply" do
+      kino = LiveCounter.new(0)
+      LiveCounter.bump(kino, 1)
+      count = LiveCounter.read_after(kino, 200)
```
jonatanklosko, line 31: This would add 0.2s to the test suit, but any timeout will test the same path :)

```suggestion
      count = LiveCounter.read_after(kino, 0)
```

jonatanklosko (approved):
Perfect, thanks!

PR #276, opened 2023-04-05
Add Kino.terminate_child/1

whatyouhide:
Without this, it took me a while to debug that a Broadway pipeline that I had started with `Kino.start_child/1` was being restarted after `Broadway.stop/1`. This was of course happening because I didn't set `restart: :transient`, but I think having `terminate_child/1` makes sense if we provide `start_child/1`.

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -543,7 +543,7 @@ defmodule Kino do
   Returns `:ok` if the child was found and terminated, or
   `{:error, :not_found}` if the child was not found.
   """
-  @doc since: "0.10.0"
+  @doc since: "0.9.1"
   @spec terminate_child(pid()) :: :ok | {:error, :not_found}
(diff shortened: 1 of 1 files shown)


lib/kino.ex
```diff
@@ -531,4 +536,16 @@ defmodule Kino do
...
     DynamicSupervisor.start_child(Kino.DynamicSupervisor, child_spec)
   end
+
+  @doc """
+  Terminates a child started with `start_child/1`.
+
+  Returns `:ok` if the child was found and terminated, or
+  `{:error, :not_found}` if the child was not found.
+  """
+  @doc since: "0.10.0"
```
josevalim, line 546:
```suggestion
  @doc since: "0.9.1"
```

josevalim (approved):

jonatanklosko (approved):

PR #283, opened 2023-05-05
Support for lazy data frames

cristineguadelupe:


josevalim (approved):
LGTM! What do we show as total runs in this case? A question mark? Or we just show nothing?

cristineguadelupe:
> LGTM! What do we show as total runs in this case? A question mark? Or we just show nothing?

A question mark and I’m disabling the `next` button when we reach a page without rows.

PR #284, opened 2023-05-12
Download exported table

cristineguadelupe:


lib/kino/table.ex
```diff
@@ -7,8 +7,9 @@ defmodule Kino.Table do
   """
 
   @type info :: %{
-          name: String.t(),
-          features: list(:refetch | :pagination | :sorting)
+          :name => String.t(),
+          :features => list(:export | :refetch | :pagination | :sorting),
+          optional(:export) => list(atom())
```
jonatanklosko, line 12:
```suggestion
          :features => list(:export | :refetch | :pagination | :sorting),
          optional(:export) => %{formats: list(atom())}
```

lib/kino/table.ex
```diff
@@ -42,6 +43,14 @@ defmodule Kino.Table do
                  total_rows: non_neg_integer() | nil
                }, state()}
 
+  @doc """
+  Exports the data for download. The returned map must contain the binary, the file extension and the mime type.
+  """
+  @callback export_data(state(), String.t()) ::
```
jonatanklosko, line 49:
```suggestion
  @doc """
  Exports the data for download.
  
  The returned map must contain the binary, the file extension and the mime type.
  """
  @callback export_data(state(), atom()) ::
```

lib/kino/table.ex
```diff
@@ -104,6 +114,19 @@ defmodule Kino.Table do
     {:noreply, broadcast_update(ctx)}
   end
 
+  def handle_event("download", %{"format" => format}, ctx) do
+    download_data = ctx.assigns.module.export_data(ctx.assigns.state, String.to_atom(format))
+
+    send_event(ctx, ctx.origin, "download_content", %{
+      "data" => download_data.data,
```
jonatanklosko, line 121: Since data can be arbitrary binary, we should use [binary payload](https://hexdocs.pm/kino/Kino.JS.Live.html#module-binary-payloads) :)

lib/kino/table.ex
```diff
@@ -89,6 +98,7 @@ defmodule Kino.Table do
     payload = %{
       name: ctx.assigns.info.name,
       features: ctx.assigns.info.features,
+      info: ctx.assigns.info,
```
jonatanklosko, line 101: We already send other things from info, so we need just the export:

```suggestion
      export: ctx.assigns.info[:export],
```

assets/data_table/main.js
```diff
@@ -101,9 +101,11 @@ function App({ ctx, data }) {
   });
 
   const hasRefetch = data.features.includes("refetch");
+  const hasExport = data.features.includes("export");
   const hasData = data.content.columns.length !== 0;
   const hasSummaries = summariesItems.length > 0;
   const hasSorting = data.features.includes("sorting");
+  const supportedFormats = data.info.export || ["csv"];
```
jonatanklosko, line 108: I think we should make the export info required if `:export` feature is enabled, this way it's clear they must support the formats they specify.

assets/data_table/main.js
```diff
@@ -421,6 +426,16 @@ function App({ ctx, data }) {
     });
   }, []);
 
+  useEffect(() => {
```
jonatanklosko, line 429: We can have all subscriptions in the same useEffect

```suggestion
    });

```

jonatanklosko (approved):

assets/data_table/main.js
```diff
@@ -421,6 +426,16 @@ function App({ ctx, data }) {
...
+    ctx.handleEvent("download_content", ([info, arrayBuffer]) => {
+      const blob = new Blob([arrayBuffer], { type: info.type });
+      const link = document.createElement("a");
+      link.href = window.URL.createObjectURL(blob);
+      link.download = `${info.filename}-${+new Date()}.${info.format}`;
```
jonatanklosko, line 434: Ah, I'd make extension include the dot as in `.csv`.

PR #291, opened 2023-06-21
Add datetime, time and date inputs

jannikbecher:


jannikbecher:
livebook-dev/livebook/pull/2002

josevalim (approved):
LGTM, we only need to address the TODOs (if @jonatanklosko agrees).

lib/kino/output.ex
```diff
@@ -297,6 +297,61 @@ defmodule Kino.Output do
...
+          | %{
+              type: :datetime,
+              ref: ref(),
+              id: input_id(),
+              label: String.t(),
+              default: NaiveDateTime.t(),
+              destination: Process.dest(),
+              min: NaiveDateTime.t(),
+              max: NaiveDateTime.t(),
+              step: number() | :any
```
jonatanklosko, line 309: I think we don't need really any?

lib/kino/input.ex
```diff
@@ -257,6 +257,218 @@ defmodule Kino.Input do
...
+
+    * `:max` - the maximum datetime value
+
+    * `:step` - the select datetime interval
+  """
+  @spec datetime(String.t(), keyword()) :: t()
+  def datetime(label, opts \\ []) when is_binary(label) and is_list(opts) do
+    min = Keyword.get(opts, :min, nil)
+    max = Keyword.get(opts, :max, nil)
+    step = Keyword.get(opts, :step, 1)
```
jonatanklosko, line 278: Actually, we need to be careful with step because it determines how precise is the input field and we must make sure if the step is like `60` (seconds), we don't use arbitrary value for the seconds when parsing the input value.

Honestly, I would rather skip `:step` for now, use the default step (60 for date and datetime), make sure we don't include milliseconds in the date structs and that seconds are zeroed out. If needed we can add support for seconds later, perhaps with a more strict option like `:precision`.

lib/kino/input.ex
```diff
@@ -257,6 +257,273 @@ defmodule Kino.Input do
...
+  ## Options
+
+    * `:default` - the initial input value. Defaults to the
+      current time
+
+    * `:min` - the minimum datetime value
+
+    * `:max` - the maximum datetime value
+
+    * `:step` - the select datetime interval
```
jonatanklosko, line 272:
```suggestion
```

lib/kino/input.ex
```diff
@@ -257,6 +257,273 @@ defmodule Kino.Input do
...
+            "expected :default to be smaller than :max, got: #{inspect(default)}"
+    end
+
+    new(%{
+      type: :datetime,
+      label: label,
+      default: default,
+      min: min,
+      max: max,
+      step: 60
```
jonatanklosko, line 307: Let's skip step entirely

```suggestion
```

lib/kino/input.ex
```diff
@@ -257,6 +257,273 @@ defmodule Kino.Input do
...
+    })
+  end
+
+  defp truncate_time(nil), do: nil
+
+  defp truncate_time(time) do
+    time
+    |> Time.truncate(:second)
+    |> Map.update!(:second, fn _ -> 0 end)
```
jonatanklosko, line 370:
```suggestion
    |> Map.replace!(:second, 0)
```

PR #294, opened 2023-07-01
Add support for Kino.JS.Live as event source

### LINKED ISSUE #290: Add listening API to Kino.JS.Live
I think we could have an API in `Kino.JS.Live` that would stream events, and we would implement stream + enumerable for `Kino.JS.Live` itself.

_Originally posted by @jonatanklosko in https://github.com/livebook-dev/livebook/discussions/1998#discussioncomment-6235988_

jonatanklosko:
Closes #290.


lib/kino/control.ex
```diff
@@ -307,7 +311,10 @@ defmodule Kino.Control do
   def stream(sources) when is_list(sources) do
     for source <- sources, do: assert_stream_source!(source)
 
-    tagged_topics = for %{attrs: %{ref: ref}} <- sources, do: {nil, ref}
+    tagged_topics =
+      for(%{attrs: %{ref: ref}} <- sources, do: {nil, ref}) ++
+        for(%Kino.JS.Live{ref: ref} <- sources, do: {nil, ref})
```
josevalim, line 316: This will change the order they are streamed. Could that be an issue? Perhaps a single `for` would be better?

test/kino/control_test.exs
```diff
@@ -109,6 +109,19 @@ defmodule Kino.ControlTest do
...
       assert events == [%{origin: "client1"}]
     end
+
+    test "supports Kino.JS.Live" do
+      kino = Kino.TestModules.LiveCounter.new(0)
+
+      spawn(fn ->
+        Process.sleep(1)
+        Kino.TestModules.LiveCounter.bump(kino, 1)
+        Kino.TestModules.LiveCounter.bump(kino, 2)
```
josevalim, line 119: It is best to write a busy loop that emits events every 10 miliseconds and then you assume two are received. Otherwise there is  no guarantee the stream will start before you emit them. Then you kill the process at the end of the test. :)

lib/kino/control.ex
jonatanklosko: What do you mean? We just subscribe to each `ref`, the order is determined by the time each event is published.

josevalim: Ok, so looks good to me. I would still do a single pass (especially because the code is already duplicate), but not a blocker. :)

test/kino/control_test.exs
josevalim: Something also nice to test is that after we take 2, nothing leaks to the inbox, so we can do an `assert_receive`. Once we have a more robust test, we can ship it!

josevalim (approved):

test/kino/control_test.exs
jonatanklosko: Smart! I updated all similar tests with a background loop. For the leak I assume you meant `refute_receive`, or do you have an alternative in mind?

lib/kino/control.ex
jonatanklosko: I updated to a single pass including validation :)

test/kino/control_test.exs
josevalim: refute_receive is excellent. You should prefer refute_received whenever possible (to avoid waiting) but in this case waiting a bit doesn't hurt (you can wait twice the time of the busy loop)

josevalim (approved):

test/kino/control_test.exs
```diff
@@ -79,16 +79,17 @@ defmodule Kino.ControlTest do
...
         send(button.attrs.destination, {:event, button.attrs.ref, info})
-        send(button.attrs.destination, {:event, button.attrs.ref, info})
       end)
 
       events = button |> Kino.Control.stream() |> Enum.take(2)
 
       assert events == [%{origin: "client1"}, %{origin: "client1"}]
+
+      # Assert that nothing leaks to the inbox
+      refute_receive _, 10
```
jonatanklosko, line 92:
```suggestion
      refute_receive _, 2
```

test/kino/control_test.exs
```diff
@@ -99,22 +100,41 @@ defmodule Kino.ControlTest do
...
+
+      background_tick(fn ->
+        Kino.TestModules.LiveCounter.bump(kino, 1)
+      end)
+
+      events = kino |> Kino.Control.stream() |> Enum.take(2)
+      assert events == [%{event: :bump, by: 1}, %{event: :bump, by: 1}]
+
+      # Assert that nothing leaks to the inbox
+      refute_receive _, 10
```
jonatanklosko, line 130:
```suggestion
      refute_receive _, 2
```

PR #299, opened 2023-07-05
Add Kino.Hub.app_info/0

jonatanklosko:
Adds function to access information about the running app, in particular, when running with Teams hub it shows the user that started the session, useful for auditing purposes.

### DIFF
--- a/lib/kino/bridge.ex
+++ b/lib/kino/bridge.ex
@@ -177,6 +177,14 @@ defmodule Kino.Bridge do
   end
 
   @doc """
+  Returns information about the running app.
+  """
+  @spec get_app_info() :: {:ok, map()} | {:error, request_error()}
+  def get_app_info() do
+    with {:ok, reply} <- io_request(:livebook_get_app_info), do: reply
+  end
+
+  @doc """
   Checks if the caller is running within Livebook context (group leader).
   """
   @spec within_livebook?() :: boolean()
--- a/lib/kino/hub.ex
+++ b/lib/kino/hub.ex
@@ -1,0 +1,33 @@
+defmodule Kino.Hub do
+  @moduledoc """
+  Functions related to hub integrations and Livebook apps.
+  """
+
+  @type app_info ::
+          %{type: :single}
+          | %{:type => :multi, optional(:creator) => user_info()}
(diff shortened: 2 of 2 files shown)


lib/kino/hub.ex
```diff
@@ -0,0 +1,33 @@
...
+  @type app_info ::
+          %{type: :single}
+          | %{:type => :multi, optional(:creator) => user_info()}
+          | %{type: :none}
+
+  @type user_info :: %{
+          id: String.t(),
+          name: String.t() | nil,
+          email: String.t() | nil,
+          source: atom()
```
jonatanklosko, line 15: I didn't make it more specific, because then it would be invalidated every time we add a new provider on Livebook side.

lib/kino/hub.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Kino.Hub do
+  @moduledoc """
+  Functions related to hub integrations and Livebook apps.
+  """
+
+  @type app_info ::
+          %{type: :single}
+          | %{:type => :multi, optional(:creator) => user_info()}
```
jonatanklosko, line 8: Also, my preference would be towards `:started_by`, since creator *could* perhaps be confused for the notebook "author".

josevalim: I am happy with `:started_by`.

lib/kino/bridge.ex
```diff
@@ -176,6 +176,14 @@ defmodule Kino.Bridge do
     end
   end
 
+  @doc """
+  Returns information about the running app.
+  """
+  @spec get_app_info() :: {:ok, map()} | {:error, request_error()}
+  def get_app_info() do
+    with {:ok, reply} <- io_request(:livebook_get_app_info), do: reply
```
josevalim, line 184: Should we return `%{type: :none}` if the `io_request` fails instead? This way we don't need to ask users to unpack the OK tuple.

jonatanklosko: `Kino.Bridge` is lower level, we have the fallback in `Kino.Hub.app_info`.

josevalim (approved):

PR #300, opened 2023-07-07
Add support for accessing notebook files

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/2044.

### DIFF
--- a/lib/kino/fs.ex
+++ b/lib/kino/fs.ex
@@ -1,4 +1,4 @@
-defmodule Kino.Files do
+defmodule Kino.FS do
   @moduledoc """
   Provides access to notebook files.
   """
--- a/lib/kino/fs.ex
+++ b/lib/kino/fs.ex
@@ -3,6 +3,19 @@ defmodule Kino.FS do
   Provides access to notebook files.
   """
 
+  defmodule ForbiddenError do
+    @moduledoc """
+    Exception raised when access to a notebook file is forbidden.
+    """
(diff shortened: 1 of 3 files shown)


josevalim (approved):

lib/kino/fs.ex
```diff
@@ -0,0 +1,67 @@
...
+               url: String.t()
+             }
+           | %{
+               type: :s3,
+               bucket_url: String.t(),
+               region: String.t(),
+               access_key_id: String.t(),
+               secret_access_key: String.t(),
+               key: String.t()
+             }
```
josevalim, line 51: Just an update: we started working on the `FSS` library. However, I believe only `Kino` should depend on `FSS`, because there is no guarantee Livebook and Kino will have the same `FSS` version. So Kino.Bridge can keep on using the format above and the function below will pattern match and convert it to the relevant FSS entries.

FSS will look like this:

    %FSS.S3.Entry{
      url: string,
      config: %FSS.S3.Config{
        access_key_id: ...
        ...
      }
    }

    %FSS.Local.Entry{
      path: string
    }

    %FSS.HTTP.Entry{
      url: string,
      config: %FSS.HTTP.Config{
        headers: [...]
      }
    }

jonatanklosko: Sounds good!

jonatanklosko:
Closing in favour of #319.

PR #306, opened 2023-07-19
Return pid when calling Kino.listen

### LINKED ISSUE #305: Why Kino.listen doesn't return a pid?
https://github.com/livebook-dev/kino/blob/78e58cfc4d8d4a7018644c55f9ba06b9100ff3ce/lib/kino.ex#L419-L428

Is there a specific reason why :ok is returned instead of the pid?
It would be nice to be able to do something like this:

```elixir
job_pid = Kino.listen(5000, fn i -> IO.puts(i) end)

Kino.Control.button("Kill job")
|> Kino.render()
|> Kino.listen(fn _ -> Process.exit(job_pid, :kill) end)
```

jannikbecher:
Closes #305.

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -320,7 +320,9 @@ defmodule Kino do
   end
 
   @doc ~S"""
-  Consumes a stream with `fun` without blocking execution.
+  Starts a process that consumes a stream with `fun` without blocking execution
+  and returns the PID of the started process.
+  The process can be terminated with `Kino.terminate_child/1`.
 
   Note that events are processed by `fun` sequentially. If you want
   to process them concurrently, use `async_listen/2`.
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -321,6 +321,7 @@ defmodule Kino do
 
   @doc ~S"""
   Starts a process that consumes a stream with `fun` without blocking execution.
+
   It returns the PID of the started process. The process can be terminated
   with `Kino.terminate_child/1`.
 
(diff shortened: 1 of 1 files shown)


lib/kino.ex
```diff
@@ -356,7 +356,7 @@ defmodule Kino do
       Kino.listen(100, fn i -> IO.puts("Ping #{i}") end)
 
   """
-  @spec listen(Enumerable.t() | pos_integer(), (term() -> any())) :: :ok
+  @spec listen(Enumerable.t() | pos_integer(), (term() -> any())) :: pid()
```
josevalim, line 359: Should we document it returns a `pid` and you can stop it with Kino.terminate?

josevalim (approved):

lib/kino.ex
```diff
@@ -320,7 +320,10 @@ defmodule Kino do
   end
 
   @doc ~S"""
-  Consumes a stream with `fun` without blocking execution.
+  Starts a process that consumes a stream with `fun` without blocking execution.
+
```
josevalim, line 324:
```suggestion

```

lib/kino.ex
```diff
@@ -320,7 +320,9 @@ defmodule Kino do
   end
 
   @doc ~S"""
-  Consumes a stream with `fun` without blocking execution.
+  Starts a process that consumes a stream with `fun` without blocking execution.
+  It returns the PID of the started process. The process can be terminated
```
jonatanklosko, line 324:
```suggestion
  Starts a process that consumes a stream with `fun` without blocking execution.

  It returns the PID of the started process. The process can be terminated
```

PR #310, opened 2023-07-31
Add kino shorts

### LINKED ISSUE #256: Add Kino.Shortcuts (or Kino.Shorts if we want to be cheeky)
A module we can import with convenience functions to building text, markdown, mermaid, image, tree, frame, etc. This is so we can write:

    text("Hello")

Instead of:

    Kino.Text.new("hello")

josevalim:
Closes #256

### DIFF
--- a/lib/kino/shorts.ex
+++ b/lib/kino/shorts.ex
@@ -47,6 +47,43 @@ defmodule Kino.Shorts do
   @spec image(binary(), Kino.Image.common_image_type() | Kino.Image.mime_type()) :: Kino.Image.t()
   def image(content, type), do: Kino.Image.new(content, type)
 
+  @doc """
+  Renders a file download button.
+
+  The given function is invoked to generate the file content whenever
+  a download is requested.
+
+  It is a wrapper around `Kino.Download.new/2`.
+
+  ## Options
+
+    * `:filename` - the default filename suggested for download.
+      Defaults to `"download"`
+
+    * `:label` - the button text. Defaults to the value of `:filename`
+      if present and `"Download"` otherwise
+
+  ## Examples
+
+      download(fn ->
+        "Example text"
+      end)
+
+      download(
+        fn -> Jason.encode!(%{"foo" => "bar"}) end,
+        filename: "data.json"
+      )
+
+      download(
+        fn -> <<0, 1>> end,
+        filename: "data.bin",
+        label: "Binary data"
+      )
+
+  """
+  @spec download((-> binary()), keyword()) :: Kino.Download.t()
+  def download(content_fun, opts \\ []), do: Kino.Download.new(content_fun, opts)
+
   @doc ~S'''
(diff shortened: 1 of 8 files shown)


josevalim:
Some notes:

1. I did not include `Kino.Process` as it is too specific

2. I also did not include `Kino.Control`. It would be ok to use them with `Kino.listen` but `Kino.Control.form` would still require `Kino.Input`, so I kept them as is

josevalim:
Some basic smoke tests are missing but this should be the bulk of the API.

lib/kino/shorts.ex
```diff
@@ -0,0 +1,573 @@
...
+
+      import Kino.Shorts
+      text("Hello!")
+
+  """
+  @spec text(String.t()) :: Kino.Text.t()
+  def text(text), do: Kino.Text.new(text)
+
+  @doc ~S'''
+  `Kino.Mermaid` renders Mermaid graphs:
```
jonatanklosko, line 135:
```suggestion
  Renders Mermaid graphs:
```

lib/kino/shorts.ex
```diff
@@ -0,0 +1,573 @@
...
+
+      html("""
+      <h3>Look!</h3>
+
+      <p>I wrote this HTML from <strong>Kino</strong>!</p>
+      """)
+
+  '''
+  @spec html(String.t()) :: Kino.HTML.t()
+  def html(html), do: Kino.HTML.new(html)
```
jonatanklosko, line 205: Shouldn't all of them be `defdelegate`?

josevalim: the delegate version with `:as` ends up being more verbose, not worth it IMO.

lib/kino/shorts.ex
```diff
@@ -0,0 +1,573 @@
...
+      * `:pcm_f32` (default) - the PCM (32-bit float) format. Note that
+        the binary uses native system endianness. Such binary can be
+        directly converted to an `Nx` tensor, with no additional decoding
+
+      * `:wav`
+
+    * `:sampling_rate` - the sampling rate (samples per second) of
+      the audio data. Defaults to `48_000`
+
+  """
```
jonatanklosko, line 522: I'm concerned with duplicating all of the docs and options and keeping them in sync. Could we just point users towards the original docs?

lib/kino/shorts.ex
jonatanklosko: Ah it's the same number of dispatches, nevermind!

lib/kino/shorts.ex
josevalim: Yeah, I am torn. We want to make it accessible for terminal and autocompletion but in this case I think it is worth just pointing to the original docs. I will update later.

jonatanklosko (approved):
🚢

josevalim:

PR #311, opened 2023-08-05
Add Kino.Audio and Kino.Video

fazibear:
Just created an `audio` kino. 

I just starting to ask myself:

- maybe this should be implemented in livebook-dev/livebook repo like Image kino? livebook-dev/livebook#1664
- what about autoplay, loop, muted  and other options?
- what about passing external urls to src?

WDYT?

Thanks for your response!

### DIFF
--- a/lib/kino/audio.ex
+++ b/lib/kino/audio.ex
@@ -30,9 +30,9 @@ defmodule Kino.Audio do
     * `:autoplay` - whether the audio should start playing as soon as
       it is rendered. Defaults to `false`
 
-    * `loop` - whether the audio should loop. Defaults to `false`
+    * `:loop` - whether the audio should loop. Defaults to `false`
 
-    * `muted` - whether the audio should be muted. Defaults to `false`
+    * `:muted` - whether the audio should be muted. Defaults to `false`
 
   """
   @spec new(binary(), common_audio_type() | mime_type(), keyword()) :: t()
--- a/lib/kino/audio.ex
+++ b/lib/kino/audio.ex
@@ -1,0 +1,57 @@
+defmodule Kino.Audio do
+  @moduledoc """
+  A kino for rendering a binary audio.
+
+  ## Examples
+
+      content = File.read!("/path/to/audio.wav")
+      Kino.Audio.new(content, :wav)
+  """
+
+  use Kino.JS
+
+  @type t :: Kino.JS.t()
+
+  @type mime_type :: binary()
+  @type common_audio_type :: :wav | :mp3 | :mpeg | :ogg
+  @doc """
+  Creates a new kino displaying the given binary audio.
+
+  The given type be either `:wav`, `:mp3`/`:mpeg`, `:ogg`
+  or a string with audio MIME type.
+
+  """
+  @spec new(binary(), common_audio_type() | mime_type()) :: t()
+  def new(content, type) when is_binary(content) do
+    Kino.JS.new(__MODULE__, data_url(content, type))
+  end
+
+  defp data_url(content, type) do
+    base64 = Base.encode64(content)
+    "data:#{mime_type!(type)};base64,#{base64}"
+  end
+
+  defp mime_type!(:wav), do: "audio/wav"
+  defp mime_type!(:mp3), do: "audio/mpeg"
+  defp mime_type!(:mpeg), do: "audio/mpeg"
+  defp mime_type!(:ogg), do: "audio/ogg"
+  defp mime_type!("audio/" <> _ = mime_type), do: mime_type
+
+  defp mime_type!(other) do
+    raise ArgumentError,
+          "expected audio type to be either :wav, :mp3, :mpeg, :ogg, or an audio MIME type string, got: #{inspect(other)}"
+  end
+
+  asset "main.js" do
+    """
+    export function init(ctx, data) {
+      ctx.root.innerHTML = `
+        <div class="root">
+            <audio controls src="${data}" />
+        </div>
+      `;
+
+    }
+    """
+  end
+end
--- a/lib/kino/audio.ex
+++ b/lib/kino/audio.ex
@@ -8,7 +8,7 @@ defmodule Kino.Audio do
       Kino.Audio.new(content, :wav)
 
       content = File.read!("/path/to/audio.wav")
-      Kino.Audio.new(content, :wav, %{autoplay: true, loop: true})
+      Kino.Audio.new(content, :wav, autoplay: true, loop: true)
   """
 
   use Kino.JS, assets_path: "lib/assets/audio"
@@ -18,6 +18,7 @@ defmodule Kino.Audio do
 
   @type mime_type :: binary()
   @type common_audio_type :: :wav | :mp3 | :mpeg | :ogg
+
   @doc """
   Creates a new kino displaying the given binary audio.
 
@@ -25,10 +26,13 @@ defmodule Kino.Audio do
   or a string with audio MIME type.
 
   ## Options
+
     * `:autoplay` - whether the audio should start playing as soon as
       it is rendered. Defaults to `false`
+
     * `loop` - whether the audio should loop. Defaults to `false`
-    * `muted` - wheater the audio shouls be muted. Defaultsa to `false`
+
+    * `muted` - whether the audio should be muted. Defaults to `false`
 
   """
   @spec new(binary(), common_audio_type() | mime_type(), keyword()) :: t()
--- a/lib/kino/audio.ex
+++ b/lib/kino/audio.ex
@@ -6,11 +6,15 @@ defmodule Kino.Audio do
 
       content = File.read!("/path/to/audio.wav")
       Kino.Audio.new(content, :wav)
+
+      content = File.read!("/path/to/audio.wav")
+      Kino.Audio.new(content, :wav, %{autoplay: true, loop: true})
   """
 
-  use Kino.JS
+  use Kino.JS, assets_path: "lib/assets/audio"
+  use Kino.JS.Live
 
-  @type t :: Kino.JS.t()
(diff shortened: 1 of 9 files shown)


lib/kino/render.ex
```diff
@@ -90,6 +90,18 @@ defimpl Kino.Render, for: Kino.Control do
...
+  def to_livebook(audio) do
+    Kino.Output.audio(audio.content, audio.mime_type)
+  end
+end
+
+defimpl Kino.Render, for: Video do
+  def to_livebook(video) do
+    Kino.Output.video(video.content, video.mime_type)
+  end
+end
```
josevalim, line 103: Please let's wait for @jonatanklosko to confirm but I am almost sure you don't need this if using Kino.JS. Although it is unclear if we should use Kino.JS or handle it in Livebook. Why is Kino.Image handled in Livebook again?

jonatanklosko: > I am almost sure you don't need this if using Kino.JS

Yeah, the changes in `render.ex` and `output.ex` are not necessary :)

> Although it is unclear if we should use Kino.JS or handle it in Livebook.

I would do it using Kino APIs, because in the future we can extend it with streaming (`Kino.JS.Live` should have all the APIs we need) and this stays transparent to Livebook.

> Why is Kino.Image handled in Livebook again?

IIRC our reasoning was that it's common enough to justify special casing and not using an iframe.

lib/kino/video.ex
```diff
@@ -0,0 +1,58 @@
...
+  @doc """
+  Creates a new kino displaying the given binary video.
+
+  The given type be either `:mp4`, `:ogg`, `:avi`, `:wmv`, `:mov`
+  or a string with video MIME type.
+
+  """
+  @spec new(binary(), common_video_type() | mime_type()) :: t()
+  def new(content, type) when is_binary(content) do
+    Kino.JS.new(__MODULE__, data_url(content, type))
```
jonatanklosko, line 26: Encoding as data URL inflates the size. Instead we can send the raw binary:

```suggestion
    Kino.JS.new(__MODULE__, {:binary, content, %{type: type}})
```

On the client we receive it as `function init(ctx, [content, { type }]) {...}`.

To convert the raw binary to base64 on the client, you can use this:

```js
function bufferToBase64(buffer) {
  let binaryString = "";
  const bytes = new Uint8Array(buffer);
  const length = bytes.byteLength;

  for (let i = 0; i < length; i++) {
    binaryString += String.fromCharCode(bytes[i]);
  }

  return btoa(binaryString);
}
```

fazibear:
```elixir 
def new(content, type) when is_binary(content) do
  Kino.JS.new(__MODULE__, {:binary, content, %{type: mime_type!(type)}})
end
```
don't work for me geting:
```
Failed to serialize initial widget data, invalid byte 0xE6 in <<82, 73,
```

jonatanklosko: Ah sorry, we only support this for `Kino.JS.Live` indeed. That said, I would actually make it `Kino.JS.Live` that stores content in the state and sends that payload on `handle_connect`. It means we use an additional process, but the advantage is that the binary is sent over a separate channel than LV operates on. For audio/video outputs it's more likely that the content is larger, so using a separate channel is better.

lib/kino/audio.ex
```diff
@@ -0,0 +1,74 @@
...
+
+  @type mime_type :: binary()
+  @type common_audio_type :: :wav | :mp3 | :mpeg | :ogg
+  @doc """
+  Creates a new kino displaying the given binary audio.
+
+  The given type be either `:wav`, `:mp3`/`:mpeg`, `:ogg`
+  or a string with audio MIME type.
+
+  """
```
jonatanklosko, line 27:
```suggestion
  or a string with audio MIME type.

  ## Options

    * `:autoplay` - whether the audio should start playing as soon as
      it is rendered. Defaults to `false`

  """
```

lib/kino/audio.ex
```diff
@@ -0,0 +1,74 @@
...
+  @spec new(binary(), common_audio_type() | mime_type(), list()) :: t()
+  def new(content, type, opts \\ [:controls, :autoplay]) when is_binary(content) do
+    Kino.JS.Live.new(__MODULE__, %{
+      content: content,
+      type: mime_type!(type),
+      opts: Enum.join(opts, " ")
+    })
+  end
```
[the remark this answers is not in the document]
fazibear: Actually there are 5 options you can pass here: `autoplay`, `controls`,  `loop `, `muted` ,`preload`. So list a bit shorter.

lib/kino/audio.ex
```diff
@@ -0,0 +1,74 @@
...
+  defp mime_type!(other) do
+    raise ArgumentError,
+          "expected audio type to be either :wav, :mp3, :mpeg, :ogg, or an audio MIME type string, got: #{inspect(other)}"
+  end
+
+  asset "main.js" do
+    """
+    export function init(ctx, [{ type, opts }, content]) {
+      let blob = new Blob([content], {type: type})
+      let data = URL.createObjectURL(blob)
```
jonatanklosko, line 63: I'd rather use the `bufferToBase64` function I sent. I'm don't remember exactly, but I'm pretty sure there is a reason we don't convert using Blob in other places, it may have been slower. Other than Blobs should be revoked (perhaps less of an issue with iframes).

lib/kino/audio.ex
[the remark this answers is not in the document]
jonatanklosko: I don't think `preload` matters, let's support `autoplay`, `loop` and `muted` for now. I still prefer regular options, because this gives us more control, and we may have other options in the future. Also with the current implementation the user could technically interpolate anything :)

lib/kino/audio.ex
fazibear: It's ok to duplicate this function?

jonatanklosko: Yeah it's fine :)

jonatanklosko: Oh, also please define the assets in files, as we do here: https://github.com/livebook-dev/kino/blob/70fc3ad2ad8350789185d9dc7120812b7aa6fd45/lib/kino/html.ex#L32

It's fine readability-wise, but since recently Livebook may try to load assets from CDN, but that's only possible if they are defined in separate files :)

lib/kino/audio.ex
```diff
@@ -0,0 +1,78 @@
...
+  The given type be either `:wav`, `:mp3`/`:mpeg`, `:ogg`
+  or a string with audio MIME type.
+
+  ## Options
+    * `:autoplay` - whether the audio should start playing as soon as
+      it is rendered. Defaults to `false`
+    * `loop` - whether the audio should loop. Defaults to `false`
+    * `muted` - wheater the audio shouls be muted. Defaultsa to `false`
+
+  """
```
jonatanklosko, line 33:
```suggestion
  @type common_audio_type :: :wav | :mp3 | :mpeg | :ogg

  @doc """
  Creates a new kino displaying the given binary audio.

  The given type be either `:wav`, `:mp3`/`:mpeg`, `:ogg`
  or a string with audio MIME type.

  ## Options

    * `:autoplay` - whether the audio should start playing as soon as
      it is rendered. Defaults to `false`

    * `loop` - whether the audio should loop. Defaults to `false`

    * `muted` - whether the audio should be muted. Defaults to `false`

  """
```

lib/kino/video.ex
```diff
@@ -0,0 +1,80 @@
...
+  @moduledoc """
+  A kino for rendering a binary video.
+
+  ## Examples
+
+      content = File.read!("/path/to/video.mp4")
+      Kino.Video.new(content, :mp4)
+
+      content = File.read!("/path/to/video.mp4")
+      Kino.Video.new(content, :mp4, [:controls, :autoplay, :loop])
```
jonatanklosko, line 11:
```suggestion
      Kino.Video.new(content, :mp4, autoplay: true, loop: true)
```

lib/kino/video.ex
```diff
@@ -0,0 +1,80 @@
...
+  The given type be either `:mp4`, `:ogg`, `:avi`, `:wmv`, `:mov`
+  or a string with video MIME type.
+
+  ## Options
+    * `:autoplay` - whether the video should start playing as soon as
+      it is rendered. Defaults to `false`
+    * `loop` - whether the video should loop. Defaults to `false`
+    * `muted` - wheater the video shouls be muted. Defaultsa to `false`
+
+  """
```
jonatanklosko, line 34:
```suggestion
  @type common_video_type :: :mp4 | :ogg | :avi | :mwv | :mov

  @doc """
  Creates a new kino displaying the given binary video.

  The given type be either `:mp4`, `:ogg`, `:avi`, `:wmv`, `:mov`
  or a string with video MIME type.

  ## Options

    * `:autoplay` - whether the video should start playing as soon as
      it is rendered. Defaults to `false`

    * `loop` - whether the video should loop. Defaults to `false`

    * `muted` - whether the video should be muted. Defaults to `false`

  """
```

test/kino/video_test.exs
```diff
@@ -0,0 +1,31 @@
...
+                   end
+    end
+
+    test "mime type shorthand and default opts" do
+      kino = Kino.Video.new(<<>>, :mp4)
+      assert {:binary, %{type: "video/mp4", opts: "controls"}, <<>>} == connect(kino)
+    end
+
+    test "custom mime type and custom opts" do
+      kino = Kino.Video.new(<<>>, "video/h123", [loop: true])
```
jonatanklosko, line 27:
```suggestion
      kino = Kino.Video.new(<<>>, "video/h123", loop: true)
```

jonatanklosko (approved):
Thanks :)

PR #314, opened 2023-08-08
Support ANSI escape sequences in `Kino.Text`

### LINKED ISSUE #313: Add ANSI escape support to `Kino.Text`
Currently, `Kino.Text` always renders as `Kino.Output.plain_text(content)`. To render text containing ANSI escape sequences, you have to create your own wrapper struct that uses `Kino.Output.text(content)`.

Being able to use `Kino.Text` to support text with ANSI escape sequences would be a nice little improvement:

```elixir
Kino.Text.new(my_text, plain: false) # default true
# or
Kino.Text.new(my_text, ansi: true) # default false
```

zachallaun:
Closes #313

### DIFF
--- a/lib/kino/text.ex
+++ b/lib/kino/text.ex
@@ -1,28 +1,39 @@
 defmodule Kino.Text do
   @moduledoc ~S'''
-  A kino for rendering plain text content.
+  A kino for rendering text content.
 
-  For rich text use `Kino.Markdown`.
+  For rich text, use `Kino.Markdown`.
 
   ## Examples
 
       Kino.Text.new("Hello!")
 
+      [:green, "Hello!"]
+      |> IO.ANSI.format()
+      |> IO.iodata_to_binary()
+      |> Kino.Text.new(console: true)
+
   '''
 
   @enforce_keys [:content]
 
-  defstruct [:content]
+  defstruct [:content, console: false]
 
   @opaque t :: %__MODULE__{
-            content: binary()
+            content: binary(),
+            console: boolean()
           }
 
   @doc """
   Creates a new kino displaying the given text content.
+
+  ## Options
+
+    * `:console` - whether to render the text as if it were printed to
+      standard output, supporting ANSI escape codes. Defaults to `false`
   """
-  @spec new(String.t()) :: t()
-  def new(content) do
-    %__MODULE__{content: content}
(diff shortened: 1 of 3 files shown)


lib/kino/text.ex
```diff
@@ -1,28 +1,39 @@
...
+
+    * `:console` - whether to render the text as if it were printed to
+      standard output, supporting ANSI escape codes. Defaults to `false`
   """
-  @spec new(String.t()) :: t()
-  def new(content) do
-    %__MODULE__{content: content}
+  @spec new(String.t(), opts) :: t() when opts: [console: boolean()]
+  def new(content, opts \\ []) when is_binary(content) do
+    %__MODULE__{content: content, console: Keyword.get(opts, :console, false)}
```
jonatanklosko, line 37:
```suggestion
    opts = Keyword.validate!(opts, console: false)
    %__MODULE__{content: content, console: opts[:console]}
```

jonatanklosko (approved):

lib/kino/text.ex
```diff
@@ -1,28 +1,39 @@
...
+            content: binary(),
+            console: boolean()
           }
 
   @doc """
   Creates a new kino displaying the given text content.
+
+  ## Options
+
+    * `:console` - whether to render the text as if it were printed to
```
jonatanklosko, line 32: Oh actually I think `:terminal` is a better word.

test/kino/text_test.exs
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    test "outputs console text" do
+      [:red, "Hello!"]
+      |> IO.ANSI.format()
+      |> IO.iodata_to_binary()
+      |> Kino.Text.new(console: true)
+      |> Kino.render()
+
+      assert_output({:text, "\e[31mHello!\e[0m"})
```
jonatanklosko, line 20: On CI Elixir disables ANSI. We could enable it explicitly, but I don't think it's relevant for the test, the content will be whatever we pass, just need to make sure it's the right output type :)

```suggestion
      "Hello!" |> Kino.Text.new(console: true) |> Kino.render()
      assert_output({:text, "Hello!"})
```

zachallaun:
All requested changes made :)

jonatanklosko:
🐈!

PR #318, opened 2023-08-21
Add support for chunked text and markdown outputs

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/2174.


lib/kino/markdown.ex
```diff
@@ -29,23 +29,32 @@ defmodule Kino.Markdown do
...
 
   @doc """
   Creates a new kino displaying the given Markdown content.
+
+  ## Options
+
+    * `:chunk` - whether this is a part of a larger text. Adjacent chunks
+        are merged into a single text. This is useful for streaming content.
+        Defaults to `false`
+
```
josevalim, line 53:
```suggestion
      are merged into a single text. This is useful for streaming content.
      Defaults to `false`

```

josevalim (approved):
LGTM! I am not sure about the name, :streaming, :squash, :chunk, :merge are all good to me. Your call.

jonatanklosko:
> I am not sure about the name, :streaming, :squash, :chunk, :merge are all good to me. Your call.

Initially my idea was to use `:squash` and always merge with the previous text regardless its mode. But then I though that if someone renders standalone text with `Kino.Text.new("yo")`, we shouldn't squash into that text for whatever reason. When rendering we know if the first text should be mergeable or not. Either of `:chunk`, `:merge` or `:fuse` would work for me, but I like `:chunk` a bit more because it speaks to the type of text rather than action :)

PR #320, opened 2023-08-23
Update outputs format

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/2179.


lib/kino/output.ex
```diff
@@ -1,512 +1,25 @@
 defmodule Kino.Output do
-  @moduledoc """
-  A number of output formats supported by Livebook.
-  """
+  @moduledoc false
```
jonatanklosko, line 2: I didn't bother deprecating, because I'm pretty sure in practice people don't use `Kino.Output` directly. Custom `Kino.Render` implementations generally build built-in kinos and then call `Kino.Render.to_livebook/1`. I also added docs to `Kino.Render` with examples suggesting that.

josevalim (approved):

PR #321, opened 2023-08-26
Allow Kino.JS.Live outputs to be exported

jonatanklosko:
Also improves the export API for `Kino.JS`.

### DIFF
--- a/lib/kino/js.ex
+++ b/lib/kino/js.ex
@@ -202,7 +202,7 @@ defmodule Kino.JS do
 
   defstruct [:module, :ref, :export]
 
-  @opaque t :: %__MODULE__{module: module(), ref: Kino.Output.ref(), export: map()}
+  @opaque t :: %__MODULE__{module: module(), ref: Kino.Output.ref(), export: boolean()}
 
   defmacro __using__(opts) do
     quote location: :keep, bind_quoted: [opts: opts] do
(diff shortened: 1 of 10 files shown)


lib/kino/js/live.ex
```diff
@@ -322,16 +327,25 @@ defmodule Kino.JS.Live do
 
   The given `init_arg` is passed to the `init/2` callback when
   the underlying kino process is started.
+
+  ## Options
+
+    * `:export` - a function called to export the given kino to Markdown.
+      This works the same as `Kino.JS.new/3`, except the function
+      receives `t:Kino.JS.Live.Context.t/0` as an argument
```
jonatanklosko, line 335: @josevalim I made this an option instead of a callback, for two reasons:

1. It better mirrors `Kino.JS`, so should be more intuitive. We also have the option on `Kino.Table`, this way they are all consistent.

2. In some cases the original data may not be passed to the server directly (see [Kino.DataTable.new/2](https://github.com/livebook-dev/kino/blob/8dac6e2b87455f9126b8860c6c3763dadc9d136c/lib/kino/data_table.ex#L69)). Passing it to the state for the purpose of export is not different from closure, but if it's large data then it may be better to just call `inspect/1` right away and close over that.

jonatanklosko: Also, I'm not sure if should actually be passing the state to that function. If we do that, then people may make the export reflect the state (e.g. page when paginating the table). It is undesired, because (a) we wouldn't know that we should re-save when the page changes, so it would end up pretty arbitrary (b) it wouldn't reflect the actual value, that is, if the cell returns `df` and we suddenly persist 10th page, it's weird. Thoughts?

lib/kino/js.ex
```diff
@@ -457,12 +461,22 @@ defmodule Kino.JS do
...
   """
   @spec new(module(), term(), keyword()) :: t()
   def new(module, data, opts \\ []) do
+    # TODO: remove the old export options in Kino v0.14.0
     export =
       if info_string = opts[:export_info_string] do
+        IO.warn(
+          "passing :export_info_string to Kino.JS.new/3 is deprecated. Specify an :export function instead"
+        )
+
```
josevalim, line 479: Will we have to update many kinos or only Kino.VegaLite?

josevalim (approved):
LGTM! It is only a question if we want to deprecate it now or in a future release. If we deprecate it now, we will have to release new packages for Kino.VegaLite that depend on latest Kino immediately. If it is only Kino.VegaLite it is fine, but if there are more, maybe we want to wait.

lib/kino/js.ex
cristineguadelupe: `Kino.Maplibre` too!

jonatanklosko:
@josevalim I think it's fine, the new kino brings important changes like support for files and the changed output format, so pushing for newer version makes sense.

lib/kino/js.ex
jonatanklosko: The only other one I found is [kino_wardely](https://github.com/chriseyre2000/kino_wardley).

hugobarauna:
Did we update our Kinos according to this?

jonatanklosko:
Not yet. We need to update `kino_vega_lite` and `kino_maplibre`, but this will require bumping to kino `~> 0.11`, so we are waiting. We may still want to update `kino_explorer` now that we can export `Kino.JS.Live`, @josevalim?

jonatanklosko:
Actually I will add a conditional to `kino_vega_lite` and `kino_maplibre` :)

PR #322, opened 2023-08-29
Add Kino.start_child!/1

jonatanklosko:
🐈

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -451,11 +451,25 @@ defmodule Kino do
   @spec start_child!(Supervisor.child_spec() | {module(), term()} | module()) :: pid()
   def start_child!(child_spec) do
     case start_child(child_spec) do
-      {:ok, pid} -> pid
-      {:ok, pid, _info} -> pid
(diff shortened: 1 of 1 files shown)


josevalim (approved):

lib/kino.ex
```diff
@@ -445,6 +445,32 @@ defmodule Kino do
...
+    end
+  end
+
+  defp start_supervised_error({{:EXIT, reason}, info}) when is_tuple(info),
+    do: Exception.format_exit(reason)
+
+  defp start_supervised_error({reason, info}) when is_tuple(info),
+    do: Exception.format_exit(reason)
+
+  defp start_supervised_error(reason), do: Exception.format_exit(reason)
```
jonatanklosko, line 472: ExUnit does `{:start_spec, reason}`, but I think it is confusing since it doesn't really have to do with the child spec itself:

```
** (RuntimeError) failed to start child with the spec Foo.
Reason: bad child specification, got: {:already_started, #PID<0.330.0>}
```

josevalim: Your call if you want to change it here. :)

Issue #325, opened 2023-09-01
Date input default value cannot be at the extremes of the allowed range


rhcarvalho:
I'm trying to use a `Kino.Input.date` in a livebook such that I want to limit the date range using `:min` and `:max`.

For example:

```ex
Kino.Input.date("From", min: ~D[2023-08-01], max: Date.utc_today())
```

Renders a widget with allowed dates 2023-08-01 all the way to 2023-09-01 (today) -- **both ends inclusive**.

---

I would like to set a default value with `:default`:

```ex
Kino.Input.date("From", default: ~D[2023-08-01], min: ~D[2023-08-01], max: Date.utc_today())
```

but I'm getting a validation error:

```
** (ArgumentError) expected :default to be bigger than :min, got: ~D[2023-08-01]
    (kino 0.10.0) lib/kino/input.ex:399: Kino.Input.date/2
```

Same on the `:max` end of the allowed range:

```ex
Kino.Input.date("From", default: Date.utc_today(), min: ~D[2023-08-01], max: Date.utc_today())
```

```
** (ArgumentError) expected :default to be smaller than :max, got: ~D[2023-09-01]
    (kino 0.10.0) lib/kino/input.ex:404: Kino.Input.date/2
```

The relevant code is:

https://github.com/livebook-dev/kino/blob/c0167dd44714fb81744b166b22decb07db7c3bce/lib/kino/input.ex#L396-L404

I think we have an off-by-one error. To match the behavior of the allowed range when using the widget we should test `default < min` and `default > max`.

josevalim:
I believe we need to use `Date.compare`.

PR #327, opened 2023-09-02
Kino.Input error messages

rhcarvalho:
Hi there!

@jonatanklosko was faster than I in addressing #325 with #326 -- thanks for the quick fix!

Since I had some changes laying around already, I decided to post them here for feedback/discussion, I wouldn't mind if they are eventually not accepted.

## ~~Tests~~

<details>

~~I noticed `Kino.Input.date` was lacking tests, and was happy to see @jonatanklosko added one yesterday!~~

~~I had a plan to cover more of the possible cases, and so tried some form of table-testing to reduce verbosity. I wonder how others see this type of test code in Elixir, given it also adds some cognitive complexity (`unquote`, `Macro.escape`, ...).~~

~~The constraints in this case are simple enough that property-based testing would also be a fun exercise :-)~~

</details>

Update: removed tests to keep the PR more focused.

## Error messages

Two out of three messages were missing part of what caused the error, making debugging unnecessarily harder.

I tried to make the messages more obvious and easier to interpret by a human when their Livebook cell is spitting out an error on them.

Something like `expected :min to be less than :max, got: #{inspect(min)} and #{inspect(max)}` is not exactly correct, as `:min` can also be *equal to* `:max`. If I were to continue the existing mathematical pattern, I'd write:

```
expected :min to be less than or equal :max, got: #{inspect(min)} and #{inspect(max)}
```

But, then, I thought why not simply state the problem directly: `:min is greater than :max`. And, when we talk about dates and time, I think the words "before" and "after" might be a better fit (might be a bias from my Go experience, where the [`time` package uses the words "before" and "after"](https://pkg.go.dev/time#Time.Compare)).

I would love to hear what are the Elixir-way ideas here, specially if there is a well establish preference.

If my changes are welcome, I could also update the other messages (time, etc) to match.

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -277,7 +277,7 @@ defmodule Kino.Input do
 
     if min && max && NaiveDateTime.compare(min, max) == :gt do
       raise ArgumentError,
-            "expected :min to be less than :max, got: #{inspect(min)} and #{inspect(max)}"
+            "expected a non-empty range, but :min (#{inspect(min)}) is after :max (#{inspect(max)})"
     end
 
     assert_default_value!(
@@ -287,11 +287,13 @@ defmodule Kino.Input do
     )
 
     if min && default && NaiveDateTime.compare(default, min) == :lt do
-      raise ArgumentError, "expected :default to be bigger than :min, got: #{inspect(default)} "
+      raise ArgumentError,
+            "invalid :default, #{inspect(default)} is before :min (#{inspect(min)})"
     end
 
     if max && default && NaiveDateTime.compare(default, max) == :gt do
-      raise ArgumentError, "expected :default to be smaller than :max, got: #{inspect(default)}"
+      raise ArgumentError,
+            "invalid :default, #{inspect(default)} is after :max (#{inspect(max)})"
     end
 
     new(%{
@@ -333,17 +335,19 @@ defmodule Kino.Input do
 
     if min && max && Time.compare(min, max) == :gt do
       raise ArgumentError,
-            "expected :min to be less than :max, got: #{inspect(min)} and #{inspect(max)}"
+            "expected a non-empty range, but :min (#{inspect(min)}) is after :max (#{inspect(max)})"
     end
 
     assert_default_value!(default, "be %Time{} or nil", &(is_struct(&1, Time) or &1 == nil))
 
     if min && default && Time.compare(default, min) == :lt do
-      raise ArgumentError, "expected :default to be bigger than :min, got: #{inspect(default)}"
+      raise ArgumentError,
+            "invalid :default, #{inspect(default)} is before :min (#{inspect(min)})"
     end
 
     if max && default && Time.compare(default, max) == :gt do
-      raise ArgumentError, "expected :default to be smaller than :max, got: #{inspect(default)}"
+      raise ArgumentError,
+            "invalid :default, #{inspect(default)} is after :max (#{inspect(max)})"
     end
 
     new(%{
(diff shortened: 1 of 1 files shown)


lib/kino/input.ex
```diff
@@ -384,17 +384,17 @@ defmodule Kino.Input do
 
     if min && max && Date.compare(min, max) == :gt do
       raise ArgumentError,
-            "expected :min to be less than :max, got: #{inspect(min)} and #{inspect(max)}"
+            ":min = #{inspect(min)} after :max = #{inspect(max)}"
```
jonatanklosko, line 387: It is a common phrasing in Elixir, but it's fine to reword, we should be more descriptive though:

```suggestion
            "empty date range, :min (#{inspect(min)}) is after :max (#{inspect(max)})"
```

lib/kino/input.ex
```diff
@@ -384,17 +384,17 @@ defmodule Kino.Input do
...
     end
 
     assert_default_value!(default, "be %Date{} or nil", &(is_struct(&1, Date) or &1 == nil))
 
     if min && default && Date.compare(default, min) == :lt do
-      raise ArgumentError, "expected :default to be bigger than :min, got: #{inspect(default)}"
+      raise ArgumentError, ":default = #{inspect(default)} before :min = #{inspect(min)}"
```
jonatanklosko, line 393:
```suggestion
      raise ArgumentError, "invalid :default, #{inspect(default)} is before :min (#{inspect(min)})"
```

lib/kino/input.ex
```diff
@@ -384,17 +384,17 @@ defmodule Kino.Input do
...
     end
 
     if max && default && Date.compare(default, max) == :gt do
-      raise ArgumentError, "expected :default to be smaller than :max, got: #{inspect(default)}"
+      raise ArgumentError, ":default = #{inspect(default)} after :max = #{inspect(max)}"
```
jonatanklosko, line 397:
```suggestion
      raise ArgumentError, "invalid :default, #{inspect(default)} is after :max (#{inspect(max)})"
```

test/kino/input_test.exs
```diff
@@ -51,22 +51,61 @@ defmodule Kino.InputTest do
...
+      {[default: ~D[2023-09-01], min: ~D[2023-09-01], max: ~D[2023-09-01]],
+       %{default: ~D[2023-09-01], min: ~D[2023-09-01], max: ~D[2023-09-01]}}
+    ]
+
+    Enum.each(@valid_inputs, fn {opts, expected_attrs} ->
+      test "creates a new date input with opts #{inspect(opts)}" do
+        assert %{attrs: unquote(Macro.escape(Map.merge(expected_attrs, %{type: :date})))} =
+                 Kino.Input.date("Date Input", unquote(Macro.escape(opts)))
+      end
+    end)
```
jonatanklosko, line 94: There are cases where generating tests with macros makes sense, but it has its drawbacks (running a specific test, quickly identifying what failed, readability). In this case, I don't think we need the whole matrix of cases. We can test a couple edge cases with different messages, this could actually be a single test :)

rhcarvalho:
@jonatanklosko I reviewed the existing error messages and based on your feedback updated the messages for all related functions:

- `Kino.Input.utc_datetime`
- `Kino.Input.utc_time`
- `Kino.Input.date`

Since the "expected ..." format is used in many places, I kept that + your suggestions.

I dropped the tests-related commit so that this PR can be focused on a single thing.

jonatanklosko (approved):
Awesome, please sync with main and it's ready to go :D

rhcarvalho:
> please sync with main and it's ready to go :D

Done!

PR #328, opened 2023-09-04
Remote cell

cristineguadelupe:
Hey @jonatanklosko! @josevalim told me to ask you if there's a better way to give feedback on invalid codes.

### DIFF
--- a/lib/assets/remote_cell/main.js
+++ b/lib/assets/remote_cell/main.js
@@ -1,0 +1,114 @@
+import * as Vue from "https://cdn.jsdelivr.net/npm/vue@3.2.26/dist/vue.esm-browser.prod.js";
+
+export function init(ctx, payload) {
+  ctx.importCSS(
+    "https://fonts.googleapis.com/css2?family=Inter:wght@400;500&display=swap"
+  );
+  ctx.importCSS("main.css");
+
+  const BaseInput = {
+    name: "BaseInput",
+
+    props: {
+      label: {
+        type: String,
+        default: "",
+      },
+      inputClass: {
+        type: String,
+        default: "input",
+      },
+      modelValue: {
+        type: [String, Number],
+        default: "",
+      },
+      inline: {
+        type: Boolean,
+        default: false,
+      },
+      grow: {
+        type: Boolean,
+        default: false,
+      },
+      number: {
+        type: Boolean,
+        default: false,
+      },
+    },
+
+    template: `
+    <div v-bind:class="[inline ? 'inline-field' : 'field', grow ? 'grow' : '']">
+      <label v-bind:class="inline ? 'inline-input-label' : 'input-label'">
+        {{ label }}
+      </label>
+      <input
+        :value="modelValue"
+        @input="$emit('update:data', $event.target.value)"
+        v-bind="$attrs"
+        :class="[inputClass, number ? 'input-number' : '']"
+      >
+    </div>
+    `,
+  };
+
+  const app = Vue.createApp({
+    components: {
+      BaseInput: BaseInput,
+    },
+
+    template: `
+    <div class="app">
+      <form @change="handleFieldChange">
+        <div class="header">
+          <BaseInput
+            name="node"
+            label="Node"
+            v-model="fields.node"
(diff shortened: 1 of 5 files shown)


jonatanklosko:
We can define [scan_eval_result/2](https://hexdocs.pm/kino/Kino.SmartCell.html#c:scan_eval_result/2) and react to evaluation error any way we want. @josevalim is it for arbitrary errors, or do you have something specific in mind?

josevalim:
@jonatanklosko it is about the case where you have invalid Elixir code in the editor. Then `Code.string_to_quoted` will fail. I suggest to emit something like, so once evaluated it gives you the proper error, as you can see in this PR:

```
Code.string_to_quoted!("invalid code here")
```

Any other suggestions? Is there a way to attach code comments to the generated code?

cristineguadelupe:
> We can define [scan_eval_result/2](https://hexdocs.pm/kino/Kino.SmartCell.html#c:scan_eval_result/2) and react to evaluation error any way we want. @josevalim is it for arbitrary errors, or do you have something specific in mind?

Wait! I think I was a bit vague. 🙈 
The problem is to provide feedback before evaluating. While typing the code and without crashing the SmartCell

jonatanklosko:
@josevalim perhaps it would be less confusing if we generated code with explicit `raise "..."`?

> Is there a way to attach code comments to the generated code?

`def to_source(attrs)` returns string, so it can be anything we want.

josevalim:
@jonatanklosko the problem is raising the explicit syntax error that we get. What do you think about generating this:

```
# Invalid code for RPC, reproducing the error below
Code.string_to_quoted!(...)
```

?

jonatanklosko:
Sounds good!

lib/kino/remote_cell.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  defp to_quoted(%{"node" => node, "cookie" => cookie, "assign_to" => var}, {:ok, code}) do
+    var = if Kino.SmartCell.valid_variable_name?(var), do: var
+    call = build_call(code) |> build_var(var)
+
+    quote do
+      node = unquote(String.to_atom(node))
+      cookie = unquote(String.to_atom(cookie))
+      Node.set_cookie(node, cookie)
+      Node.connect(node)
```
josevalim, line 60: I think we actually don't need this call. `:erpc` connects automatically! Please verify:

```suggestion
```

lib/kino/remote_cell.ex
```diff
@@ -0,0 +1,88 @@
...
+    to_quoted(attrs, code)
+  end
+
+  defp to_quoted(%{"node" => node, "cookie" => cookie, "assign_to" => var}, {:ok, code}) do
+    var = if Kino.SmartCell.valid_variable_name?(var), do: var
+    call = build_call(code) |> build_var(var)
+
+    quote do
+      node = unquote(String.to_atom(node))
+      cookie = unquote(String.to_atom(cookie))
```
josevalim, line 58: Btw, we should allow the cookie to be a secret, but we can move that to a separate issue. :)

lib/kino/remote_cell.ex
```diff
@@ -0,0 +1,88 @@
...
+      Node.set_cookie(node, cookie)
```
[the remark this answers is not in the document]
cristineguadelupe: I think that if we do it for one, we should do it for both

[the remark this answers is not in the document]
josevalim: I don't think so because `node` we repeat (we use it more than once). We don't do so for cookie!

josevalim (approved):

lib/assets/remote_cell/main.js
```diff
@@ -0,0 +1,110 @@
...
+            :required
+          />
+          <BaseInput
+            name="assign_to"
+            label="Assign to"
+            v-model="fields.assign_to"
+            inputClass="input input--xs"
+            :inline
+          />
+      </form>
```
jonatanklosko, line 82:
```suggestion
          />
        </div>
      </form>
```

PR #336, opened 2023-09-10
Fix Kino.Input.utc_datetime opts validation

### LINKED ISSUE #335: Kino.Input.utc_datetime uses DateTime.compare to compare NaiveDateTime values
_Found while working on #327, thanks to `ElixirLS Dialyzer`._

Turns out that passing any option to `Kino.Input.utc_datetime` will cause a `FunctionClauseError (no function clause matching in DateTime.compare/2)`.

rhcarvalho:
The min, max and default options are all converted to a `NaiveDateTime`, so we cannot compare them using `DateTime`.

Fixes #335.


jonatanklosko (approved):
Nice catch, thanks!

PR #337, opened 2023-09-11
Bump versions in GitHub test workflow

rhcarvalho:
Fixes warning in recent workflow runs:

> The following actions uses node12 which is deprecated and will be forced
> to run on node16: actions/checkout@v2, actions/cache@v2.
>
> For more info:
> https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Example run: https://github.com/livebook-dev/kino/actions/runs/6142008298

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -12,18 +12,18 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: "1.15.2"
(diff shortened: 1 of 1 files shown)


jonatanklosko (approved):
Thanks!

(Apparently I didn't get notifications for the two PRs)

PR #340, opened 2023-09-13
Global attrs and global name and cookie for RemoteCell

cristineguadelupe:
I did the synchronization between secrets and plain text implicit in `Global`. That way, following our pattern of using the suffix "_secret", the SmartCell wouldn't have to worry about synchronizing it manually when using globals.
WDYT?


cristineguadelupe:
I also wondered if it wouldn't be better to have a single module for global and counter 🤔

lib/kino/smart_cell.ex
```diff
@@ -391,4 +391,15 @@ defmodule Kino.SmartCell do
     |> Inspect.Algebra.format(90)
     |> IO.iodata_to_binary()
   end
+
+  def global_attr(cell, key) do
+    case Kino.Global.lookup("#{cell}-#{key}") do
```
josevalim, line 396: The key doesn't have to be a string. I would recommend storing a tuple as key:

    {Kino.RemoteCell, :node}
    {Kino.RemoteCell, :secret}

And then, you should also store the secret as a tuple. If it is a plain cookie, store it as `{value, nil}`. If it is a cookie secret, store it as `{nil, value}`. Then you can read it as:

    {global_cookie, global_cookie_secret} = SmartCell.global_attr(@global_key, :cookie, {nil, nil})

Where the third argument is the default value (which means you need to provide different names for the function to read and the function to write).

I am also not sure if we should call it global, because it is not shared across nodes. Maybe shared_attribute? So I would have:

    Kino.SmartCell.get_shared_attribute(cell, key, default \\ nil)
    Kino.SmartCell.put_shared_attribute(cell, key, value)

Also, if we want to make it public, we would need to add docs? Otherwise we add `@doc false`. :)

jonatanklosko: I think it can stay private until we actually need it somewhere else. We don't expose the counter either, just the high-level variable generation.

lib/kino/remote_ execution_cell.ex
jonatanklosko, line 1: There's a space in the filename btw :)

jonatanklosko:
> I also wondered if it wouldn't be better to have a single module for global and counter 🤔

Sounds good, we can reuse the same ETS table :)

lib/kino/smart_cell.ex
```diff
@@ -391,4 +391,12 @@ defmodule Kino.SmartCell do
...
     |> IO.iodata_to_binary()
   end
+
+  def get_shared_attr(key) do
+    Kino.SharedAttribute.get_attr(key)
+  end
+
+  def put_shared_attr(key, value) do
+    Kino.SharedAttribute.put_attr(key, value)
+  end
```
jonatanklosko, line 401: Let's not expose, and since it's a simple delegate, we can just call these directly :)

```suggestion
```

test/kino/shared_attribute_test.exs
```diff
@@ -0,0 +1,50 @@
+defmodule Kino.SharedAttributeTets do
+  use ExUnit.Case, async: true
+
+  alias Kino.SharedAttribute
+
+  describe "next/2" do
```
jonatanklosko, line 6:
```suggestion
  describe "counter_next/2" do
```

test/kino/shared_attribute_test.exs
```diff
@@ -0,0 +1,50 @@
...
+    test "increments the counter on every call", %{test: test} do
+      SharedAttribute.counter_next(test)
+
+      assert SharedAttribute.counter_next(test) == 2
+      assert SharedAttribute.counter_next(test) == 3
+      assert SharedAttribute.counter_next(test) == 4
+    end
+  end
+
+  describe "bump/2" do
```
jonatanklosko, line 20:
```suggestion
  describe "counter_put_max/2" do
```

lib/kino/shared_attribute.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Kino.SharedAttribute do
```
jonatanklosko, line 1: `Kino.AttributeStore` perhaps?

lib/kino/shared_attribute.ex
```diff
@@ -0,0 +1,54 @@
...
+    [_, counter] =
+      :ets.update_counter(@table_name, key, [{2, -1, value, value - 1}, {2, 1}], {key, 0})
+
+    counter
+  end
+
+  @doc """
+  Puts the shared attribute value.
+  """
+  @spec put_attr(term(), term()) :: boolean()
```
jonatanklosko, line 39: `attr` -> `attribute` :)

lib/kino/shared_attribute.ex
```diff
@@ -0,0 +1,54 @@
...
+    :ets.insert(@table_name, {key, value})
+  end
+
+  @doc """
+  Returns the attribute value for a given key.
+  """
+  @spec get_attr(term()) :: term()
+  def get_attr(key) do
+    case :ets.lookup(@table_name, key) do
+      [] -> nil
```
josevalim, line 50:
```suggestion
  def get_attr(key, default \\ nil) do
    case :ets.lookup(@table_name, key) do
      [] -> default
```

lib/kino/remote_execution_cell.ex
```diff
@@ -5,16 +5,24 @@ defmodule Kino.RemoteExecutionCell do
...
+  alias Kino.SmartCell
+
   @default_code ":ok"
+  @global_key __MODULE__
+  @global_attrs ["node", "cookie", "cookie_secret"]
 
   @impl true
   def init(attrs, ctx) do
+    shared_cookie = SmartCell.get_shared_attr({@global_key, :cookie})
+    {shared_cookie, shared_cookie_secret} = if shared_cookie, do: shared_cookie, else: {nil, nil}
```
josevalim, line 17:
```suggestion
    {shared_cookie, shared_cookie_secret} =
      SmartCell.get_shared_attr({@global_key, :cookie}, {nil, nil})
```

josevalim (approved):

jonatanklosko (approved):

PR #346, opened 2023-09-21
Enable livemd export for Kino.DataTable

jonatanklosko:
🐈

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -66,7 +66,11 @@ defmodule Kino.DataTable do
         {data, meta.columns}
       end
 
-    Kino.Table.new(__MODULE__, {data_rows, data_columns, count, name, sorting_enabled})
+    inspected = inspect(tabular)
+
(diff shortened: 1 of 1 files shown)


lib/kino/data_table.ex
```diff
@@ -66,7 +66,11 @@ defmodule Kino.DataTable do
         {data, meta.columns}
       end
 
-    Kino.Table.new(__MODULE__, {data_rows, data_columns, count, name, sorting_enabled})
+    inspected = inspect(tabular)
```
josevalim, line 69: Should we inspect lazily? I guess we have to choose between keeping a reference to the inspect vs to the whole data.

jonatanklosko: For small data it doesn't matter, and if it's large data then inspect is less copying (or zero-copy if big enough).

josevalim: I was not worried about the copying but the amount of in-memory data we need to keep. I guess at best we can guess for now (and if the data is really large, inspect truncates, so that's a plus).

josevalim (approved):

PR #347, opened 2023-09-23
Add debounce option for inputs

PeterUstinox:
adds debounce option pretty much the same way as `phx-debounce` from LiveView works, except of a different default value.

This change is accompanied by a change in Livebook, see https://github.com/livebook-dev/livebook/pull/2224 for details

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -62,16 +62,15 @@ defmodule Kino.Input do
 
     * `:default` - the initial input value. Defaults to `""`
 
-    * `:debounce` - the option to modify the change event behavior.
-      Accepts an integer timeout (in milliseconds) or `"blur"`.
-      If an integer, the event emits after the specified delay.
-      If `"blur"`, it emits when the field is deselected by the user.
-      Defaults to `"blur"`
+    * `:debounce` - determines when input changes are emitted. When
+      set to `:blur`, the change propagates when the user leaves the
+      input. When set to a non-negative number of milliseconds, the
+      change propagates after the specified delay. Defaults to `:blur`
   """
   @spec text(String.t(), keyword()) :: t()
   def text(label, opts \\ []) when is_binary(label) and is_list(opts) do
     default = opts |> Keyword.get(:default, "") |> to_string()
-    debounce = Keyword.get(opts, :debounce, "blur")
+    debounce = Keyword.get(opts, :debounce, :blur)
 
     assert_debounce_value!(debounce)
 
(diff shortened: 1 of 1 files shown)


lib/kino/input.ex
```diff
@@ -61,11 +61,21 @@ defmodule Kino.Input do
   ## Options
 
     * `:default` - the initial input value. Defaults to `""`
+
+    * `:debounce` - the option to modify the change event behavior.
+      Accepts an integer timeout (in milliseconds) or `"blur"`.
+      If an integer, the event emits after the specified delay.
+      If `"blur"`, it emits when the field is deselected by the user.
+      Defaults to `"blur"`
```
jonatanklosko, line 69:
```suggestion
    * `:debounce` - determines when input changes are emitted. When
      set to `:blur`, the change propagates when the user leaves the
      input. When set to a non-negative number of milliseconds, the
      change propagates after the specified delay. Defaults to `:blur`

```

Please update the other ones accordingly :)

Also, please change `"blur"` to `:blur`.

lib/kino/input.ex
```diff
@@ -133,6 +172,13 @@ defmodule Kino.Input do
     end
   end
 
+  defp assert_debounce_value!(value) do
+    unless is_nil(value) or value == "blur" or (is_number(value) and value >= 0) do
```
jonatanklosko, line 176: We never expect `nil`

```suggestion
    unless value == :blur or (is_number(value) and value >= 0) do
```

lib/kino/input.ex
```diff
@@ -133,6 +172,13 @@ defmodule Kino.Input do
...
+      raise ArgumentError,
+            ~s/expected :debounce to be "blur" or a number greater or equal to zero, got: #{inspect(value)}/
```
jonatanklosko, line 178:
```suggestion
            ~s/expected :debounce to be :blur or a non-negative number, got: #{inspect(value)}/
```

lib/kino/input.ex
```diff
@@ -268,12 +333,19 @@ defmodule Kino.Input do
...
     * `:max` - the maximum datetime value (in UTC)
+
+    * `:debounce` - the option to modify the change event behavior.
+      Accepts an integer timeout (in milliseconds) or `"blur"`.
+      If an integer, the event emits after the specified delay.
+      If `"blur"`, it emits when the field is deselected by the user.
+      Defaults to `"blur"`
   """
   @spec utc_datetime(String.t(), keyword()) :: t()
   def utc_datetime(label, opts \\ []) when is_binary(label) and is_list(opts) do
```
jonatanklosko, line 344: This should be `date`, but let's not do this for dates for now :)

lib/kino/input.ex
PeterUstinox: 👍

lib/kino/input.ex
PeterUstinox: 👍

lib/kino/input.ex
PeterUstinox: 👍

lib/kino/input.ex
PeterUstinox: 👍

PeterUstinox:
@jonatanklosko thanks very much for reviewing! 

I just learned, I could've applied your changes via GitHub directly accrediting you. Now I did this manually and hope this is fine.
All of your suggestions were applied. I also removed the debounce from Livebook accordingly (https://github.com/livebook-dev/livebook/pull/2224)

jonatanklosko:
> I just learned, I could've applied your changes via GitHub directly, also accrediting you. Now I did this manually and hope this is fine.

No worries, it's totally fine! Code suggestions just show nice diffs, but you can apply them however is more convenient :)

jonatanklosko (approved):
Thanks :)

PR #350, opened 2023-10-03
Fix updateNodeInfo call from secret component

ghost:
~That's a necessary fix, but it doesn't solve the whole problem.
We are passing the secret's name and not its value, we don't have it here. But on the Livebook side we are not retrieving the value because there is no information that it's a secret.
Basically, we need to decide which side will be responsible for doing this.~

### DIFF
--- a/lib/kino/remote_execution_cell.ex
+++ b/lib/kino/remote_execution_cell.ex
@@ -24,7 +24,7 @@ defmodule Kino.RemoteExecutionCell do
       "assign_to" => attrs["assign_to"] || "",
       "node" => node,
       "cookie" => attrs["cookie"] || shared_cookie || "",
-      "cookie_secret" => cookie_secret,
+      "cookie_secret" => cookie_secret || "",
       "use_cookie_secret" =>
         if(shared_cookie, do: false, else: Map.get(attrs, "use_cookie_secret", true)),
       "cookie_secret_value" => cookie_secret_value


cristineguadelupe:
@hugobarauna It should work now!

lib/kino/remote_execution_cell.ex
```diff
@@ -16,15 +16,18 @@ defmodule Kino.RemoteExecutionCell do
     {shared_cookie, shared_cookie_secret} =
       AttributeStore.get_attribute({@global_key, :cookie}, {nil, nil})
 
-    node = attrs["node"] || AttributeStore.get_attribute({@global_key, :node})
+    node = attrs["node"] || AttributeStore.get_attribute({@global_key, :node}) || ""
+    cookie_secret = attrs["cookie_secret"] || shared_cookie_secret || ""
+    cookie_secret_value = System.get_env("LB_#{cookie_secret}")
```
josevalim, line 21:
```suggestion
    cookie_secret = attrs["cookie_secret"] || shared_cookie_secret
    cookie_secret_value = cookie_secret && System.get_env("LB_#{cookie_secret}")
```

lib/kino/remote_execution_cell.ex
```diff
@@ -16,15 +16,18 @@ defmodule Kino.RemoteExecutionCell do
...
 
     fields = %{
       "assign_to" => attrs["assign_to"] || "",
-      "node" => node || "",
+      "node" => node,
       "cookie" => attrs["cookie"] || shared_cookie || "",
-      "cookie_secret" => attrs["cookie_secret"] || shared_cookie_secret || "",
+      "cookie_secret" => cookie_secret,
```
josevalim, line 27: If you do the above, you may need to do this:

```suggestion
      "cookie_secret" => cookie_secret || "",
```

Not sure.

josevalim (approved):

PR #352, opened 2023-10-05
Fix input range on Safari

cristineguadelupe:
`phx-debounce="blur"` is the problem

### DIFF
--- a/lib/kino/input.ex
+++ b/lib/kino/input.ex
@@ -283,7 +283,7 @@ defmodule Kino.Input do
     max = Keyword.get(opts, :max, 100)
     step = Keyword.get(opts, :step, 1)
     default = Keyword.get(opts, :default, min)
-    # In Safari range input is blured as soon as it's clicked,
+    # In Safari range input is blurred as soon as it's clicked,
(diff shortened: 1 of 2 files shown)


lib/kino/input.ex
```diff
@@ -283,7 +283,7 @@ defmodule Kino.Input do
     max = Keyword.get(opts, :max, 100)
     step = Keyword.get(opts, :step, 1)
     default = Keyword.get(opts, :default, min)
-    debounce = Keyword.get(opts, :debounce, :blur)
+    debounce = Keyword.get(opts, :debounce, 300)
```
jonatanklosko, line 286:
```suggestion
    debounce = Keyword.get(opts, :debounce, 250)
```

jonatanklosko (commented):
Good catch, it looks like safari doesn't support blur (or rather it blurs the input as soon as it's focused). So we should remove `:blur` from supported values for the range input. We have to:

* update the the `:debounce` docs and mention the new default value
* change the validation for range input to only allow integers

There's also one change on the Livebook side, but I will do it :)

jonatanklosko (approved):

PR #359, opened 2023-10-13
Remote cell - help box

cristineguadelupe:


lib/assets/remote_execution_cell/main.js
```diff
@@ -211,6 +214,9 @@ export function init(ctx, payload) {
             inputClass="input input--xs"
             :inline
           />
+          <span class="tooltip left icon-button grow" :data-tooltip="'The Erlang/OTP version of both nodes need to match.'">
```
jonatanklosko, line 217: For consistency we probably want the same collapsible help section as in Neural network cell and SQL cell (rather than a tooltip). And we may want to add more information here in the future.

jonatanklosko (approved):

PR #363, opened 2023-11-08
Make remote execution cell work regardless of Elixir and OTP versions

jonatanklosko:
🐈‍⬛


lib/kino/rpc.ex
```diff
@@ -0,0 +1,126 @@
...
+  Returns the value returned from evaluation.
+
+  The code is analyzed for variable references, they are automatically
+  extracted from the caller binding and passed to the evaluation. This
+  means that the evaluated string actually has closure semantics.
+
+  The code is parsed and expanded on the remote node. Also, errors
+  and exists are captured and propagated to the caller.
+
+  See `Code.eval_string/3` for a description of `binding` and `opts`.
```
jonatanklosko, line 20:
```suggestion
  See `Code.eval_string/3` for available `opts`.
```

lib/kino/rpc.ex
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  defp used_var_names(string, env) do
+    case silent_string_to_quoted(string) do
+      {:ok, ast} ->
+        names = for {name, nil} <- Macro.Env.vars(env), into: %{}, do: {name, nil}
+
+        ast
+        |> Macro.prewalk(MapSet.new(), fn
+          {name, _, nil} = node, acc when is_atom(name) and is_map_key(names, name) ->
```
jonatanklosko, line 48:
```suggestion
          {name, _, nil} = node, acc when is_map_key(names, name) ->
```

mrluc:
Haha, seems like an obvious way to fix this in retrospect! There's elixir on the other side, so do version-safe rpc of just Code.eval_string, on a string of user's code, instrumented.

```elixir
...
    {nil, binding} =
      :erpc.call(node, Code, :eval_string, [   #<-- this line right here
        eval_string,
        [input: {string, binding, opts}],
        [file: @relative_file, line: line]
      ])

    {result, diagnostics} = binding[:output]

    for diagnostic <- diagnostics do
      Code.print_diagnostic(diagnostic)
    end

    case result do
      {:ok, value} ->
        value
...
```

DECE

lib/kino/rpc.ex
```diff
@@ -0,0 +1,133 @@
...
+  means that the evaluated string actually has closure semantics.
+
+  The code is parsed and expanded on the remote node. Also, errors
+  and exists are captured and propagated to the caller.
+
+  See `Code.eval_string/3` for available `opts`.
+  """
+  defmacro eval_string(node, string, opts \\ []) do
+    unless is_binary(string) do
+      raise ArgumentError, "Kino.RPC.eval_string/3 expects a string literal as an argument"
```
josevalim, line 24:
```suggestion
      raise ArgumentError, "Kino.RPC.eval_string/3 expects a string literal as the second argument"
```

lib/kino/rpc.ex
```diff
@@ -0,0 +1,133 @@
...
+        unquote(node),
+        unquote(string),
+        unquote(binding),
+        unquote(opts)
+      )
+    end
+  end
+
+  defp used_var_names(string, env) do
+    case silent_string_to_quoted(string) do
```
josevalim, line 42:
```suggestion
    case Code.string_to_quoted(string, emit_warnings: false, warn_on_unnecessary_quotes: false) do
```

This will be enough to silence it.

josevalim: We can add a TODO to keep only emit_warnings on v1.16+

josevalim (approved):

lib/kino/rpc.ex
```diff
@@ -0,0 +1,133 @@
...
+      {:ok, ast} ->
+        # This is a simple heuristic, we traverse the unexpanded AST
+        # and look for any variable node. This means we may have false
+        # positives if there are macros, but in our use case this is
+        # acceptable. We may also have false negatives in very specific
+        # edge cases, such as calling `binding()`, but these are even
+        # more unlikely.
+
+        names = for {name, nil} <- Macro.Env.vars(env), into: %{}, do: {name, nil}
```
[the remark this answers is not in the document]
jonatanklosko: LOL!

PR #366, opened 2023-11-14
Fix code with interpolation in remote execution smart cell

### LINKED ISSUE #365: Bug happening when running remote execution smart cell
Running this code works:

CleanShot 2023-11-14 at 16 02 21

But running that doesn't work:

CleanShot 2023-11-14 at 16 04 34

Here's the code generated by the smart cell, when the code doesn't work:

```elixir
require Kino.RPC
node = String.to_atom(System.fetch_env!("LB_TEAMS_PROD_NODE"))
Node.set_cookie(node, String.to_atom(System.fetch_env!("LB_TEAMS_PROD_COOKIE")))

Kino.RPC.eval_string(
  node,
  """
  Hub.Orgs.Org.changeset(org, %{name: "#{org.name}"})
  |> Hub.Repo.update()
  """,
  file: __ENV__.file
)
```

That said, when the bug happens, even I run the generated code by the smart cell, when the code worked, now it doesn't:

CleanShot 2023-11-14 at 16 06 26

jonatanklosko:
Closes #365.

### DIFF
--- a/lib/kino/rpc.ex
+++ b/lib/kino/rpc.ex
@@ -20,10 +20,18 @@ defmodule Kino.RPC do
   See `Code.eval_string/3` for available `opts`.
   """
   defmacro eval_string(node, string, opts \\ []) do
-    unless is_binary(string) do
-      raise ArgumentError,
-            "Kino.RPC.eval_string/3 expects a string literal as the second argument"
-    end
+    string =
+      case string do
+        string when is_binary(string) ->
+          string
+
+        {sigil, _, [{:<<>>, _, [string]}, []]} when sigil in [:sigil_s, :sigil_S] ->
+          string
+
+        _other ->
+          raise ArgumentError,
+                "Kino.RPC.eval_string/3 expects a string literal as the second argument"
+      end
 
     used_var_names = used_var_names(string, __CALLER__)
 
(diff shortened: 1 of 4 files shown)


lib/kino/rpc.ex
```diff
@@ -20,10 +20,18 @@ defmodule Kino.RPC do
...
-    unless is_binary(string) do
-      raise ArgumentError,
-            "Kino.RPC.eval_string/3 expects a string literal as the second argument"
-    end
+    string =
+      case string do
+        string when is_binary(string) ->
+          string
+
+        {sigil, _, [{:<<>>, _, [string]}, []]} when sigil in [:sigil_s, :sigil_S] ->
```
jonatanklosko, line 28: @josevalim please double-check :)

lib/kino/rpc.ex
```diff
@@ -20,10 +20,18 @@ defmodule Kino.RPC do
   See `Code.eval_string/3` for available `opts`.
   """
   defmacro eval_string(node, string, opts \\ []) do
-    unless is_binary(string) do
-      raise ArgumentError,
-            "Kino.RPC.eval_string/3 expects a string literal as the second argument"
-    end
+    string =
+      case string do
```
josevalim, line 24: Maybe call `Macro.expand(string, __CALLER__)` and check if it returns a string?

josevalim (approved):

PR #371, opened 2023-11-29
Fix Kino.DataTable to handle integer lists that are not valid charlists

### LINKED ISSUE #370: Postgre columns of type array can crash SQL smart cells.
I tried to fetch data from a table which happened to contain a column of type `BIGINT[]`.

This can trigger the following error:

```
18:40:37.572 [error] GenServer #PID<0.298.0> terminating
** (UnicodeConversionError) invalid code point 11933055
    (elixir 1.15.6) lib/list.ex:1103: List.to_string/1
    (kino 0.11.3) lib/kino/data_table.ex:249: Kino.DataTable.value_to_string/1
    (elixir 1.15.6) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
```
Note that 11933055 (from `invalid code point 11933055`) is an element in the array.

Steps to reproduce are detailed in the attached livebook. (Had to change the extension to `.md` or else GitHub would not allow me to attach the file.)

[arrays-not-supported.livemd.md](https://github.com/livebook-dev/kino/files/13493725/arrays-not-supported.livemd.md)

jonatanklosko:
Closes #370.

### DIFF
--- a/lib/kino/data_table.ex
+++ b/lib/kino/data_table.ex
@@ -248,7 +248,7 @@ defmodule Kino.DataTable do
     try do
       List.to_string(value)
     rescue
-      ArgumentError ->
+      _error ->
         inspect(value)
     end
   end
--- a/test/kino/data_table_test.exs
+++ b/test/kino/data_table_test.exs
@@ -102,7 +102,8 @@ defmodule Kino.DataTableTest do
     entries = [
       [
         {"a", [1, "b"]},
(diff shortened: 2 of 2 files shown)


josevalim:
Maybe we should use List.ascii_printable?, similar to inspect itself? I think it is best to not assume most lists are char lists.

jonatanklosko:
Good call, updated in 33bc23dae0893a07ac5f7dcfb9fd6836c18e7bb3!

PR #376, opened 2024-01-03
Fix list cell bugs - text when clicked and appropriate list icon

lkarthee:
When a list cell is clicked - it behaves weird, it appends list into a single word. 

https://github.com/livebook-dev/kino/assets/2712395/bbe92939-a2c2-4dda-98a4-111956974dd0

After fix

image

Also the list column was using an `GridColumnIcon.HeaderNumber` icon when `GridColumnIcon.HeaderArray` icon is more appropriate.


lkarthee:
I have done an `npm  install --force` as normal install was failing due to incompatible react dependency. There after did a `npm run build` and tested the cell. Hopefully I have not missed anything.

cristineguadelupe:
Thanks!

cristineguadelupe (approved):

PR #384, opened 2024-01-21
Fix table download to match the spec

cristineguadelupe:
Only today did I realize that I wasn't following the spec correctly.
I've fixed both sides and we can release simultaneously if you think it's worth it


jonatanklosko (approved):
Sounds good! You can bump the dependency on kino once released to ensure compatibility :)

PR #391, opened 2024-01-31
Make the smart cell editor source explicitly managed

jonatanklosko:
This is a breaking change, but I think it's better to do it sooner than later. Currently the user specifies under which `attr` the editor source should be placed and it is only available inside the `to_source(attrs)` callback. This PR makes the flow more explicit, we expose a new callback `handle_editor_change(source, ctx)` and it is up to the user to keep it in assigns and put in `attrs`. This could be used to update the UI based on the editor source (for example to parse the code and show errors).

Note that this is a breaking change only in that the smart cell implementation needs to be updated (a simple change). It doesn't change anything on Livebook side, and smart cell serialization is also not affected (provided the same attribute is used).

### DIFF
--- a/lib/kino/remote_execution_cell.ex
+++ b/lib/kino/remote_execution_cell.ex
@@ -41,13 +41,14 @@ defmodule Kino.RemoteExecutionCell do
 
     intellisense_node = intellisense_node(fields)
 
-    ctx = assign(ctx, fields: fields)
+    code = attrs["code"] || @default_code
+
+    ctx = assign(ctx, fields: fields, code: code)
 
     {:ok, ctx,
      editor: [
-       attribute: "code",
+       source: code,
        language: "elixir",
-       default_source: @default_code,
        intellisense_node: intellisense_node
      ]}
   end
@@ -98,8 +99,13 @@ defmodule Kino.RemoteExecutionCell do
   end
 
   @impl true
+  def handle_editor_change(source, ctx) do
+    {:ok, assign(ctx, code: source)}
+  end
+
+  @impl true
   def to_attrs(ctx) do
-    ctx.assigns.fields
+    Map.put(ctx.assigns.fields, "code", ctx.assigns.code)
   end
 
   @impl true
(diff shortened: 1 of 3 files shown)


jonatanklosko:
@josevalim `kino_db` has `{:kino, "~> 0.7"}`, I would release one version that depends on <= 0.12, and then another that requires 0.13 (once released). Does that make sense?

josevalim:
@jonatanklosko what if we detect if `handle_editor_change` is defined and, if not, we fallback to old behaviour?

josevalim:
Just temporarily as part of the deprecation period.

jonatanklosko:
We could detect based on whether they configure using `:source` or `:attribute`. But I'm not sure it's worth supporting both behaviours, given it's a fairly rare use case and a pretty simple change to adapt, especially if 0.13 is going to be the last one before 1.0. The `kino_db` case is still the same either way.

josevalim:
Having it, even for one version, will aid with breaking changes, so I think we should do it. :)

jonatanklosko:
Ok, so the current `kino_db` will work under the deprecation and we only release next version depending on `~> 0.13`, and do 1.0 once we have kino 1.0?

josevalim:
@jonatanklosko yes!

josevalim:
Btw, do you think it is worth a new callback? Could we just send it as a regular event (perhaps make the event name configurable under the new API?)?

josevalim:
Well, we need the new callback for the deprecation, so nevermind. :D

jonatanklosko:
We will deprecate based on the attributes, previously `[editor: [attribute: "..."]]` was required and now `[editor: [source: "..."]]` is required.

The main difference with the callback is that we can error sooner if the user enables editor and doesn't define the callback, but either works for me, so whichever you like more :)

josevalim:
@jonatanklosko I have no preference. Both options are fine. Given that the smart cell is mostly optional callbacks, keeping it as an optional callback is also easier to discover, so it probably justifies it. But it is your call. :)

jonatanklosko:
Callback it is :) Updated!

josevalim (approved):

PR #394, opened 2024-02-06
Use shared bundling setup for all assets

jonatanklosko:
I created a small package wrapping common setup for bundling kino assets ([@livebook/kino-bundler](https://github.com/livebook-dev/kino-bundler?tab=readme-ov-file)) and we now use it for all assets in this repo.

Notably, the package brings Tailwind configured to match Livebook, which means that Kinos can easily use the Livebook palette without copying individual colors, which we used to do over and over.

Given that kino components are quite small, I don't think it's worth maintaining a package with shared UI components. The main pain point so far was that whenever we copied components, we needed to copy the relevant CSS classes and that could quickly lead to unnecessary or missing classes and was messy overall. Now that each component is properly encapsulated, copying a couple of them around should be manageable, while being easy to change if needed.

I migrated the Remote execution cell to React, so there is no more Vue in this repo. It would be nice to migrate the other smart cells accordingly, but it's not a high priority, the important part is that we have a better reference implementation.

The current setup of bundler and React also improves dev experience, now `npm run dev` includes dev warnings and source maps.


cristineguadelupe:
Nice! 😊

josevalim (approved):

cristineguadelupe (approved):

PR #399, opened 2024-02-29
Add Kino.recompile/0

jonatanklosko:
🐈


hugobarauna:
Out of curiosity: why is this in Kino?

PR #402, opened 2024-03-22
Add doc examples about the chunk option of Kino.Text.new/2 and Kino.Markdown.new/2

hugobarauna:
A user asked in the Slack channel about how to use the chunk option. I noticed there were no examples in the docs. So I copied the examples [from the original PR adding that option](https://github.com/livebook-dev/livebook/pull/2174) to the docs.

### DIFF
--- a/lib/kino/markdown.ex
+++ b/lib/kino/markdown.ex
@@ -51,6 +51,29 @@ defmodule Kino.Markdown do
       are merged into a single text. This is useful for streaming content.
       Defaults to `false`
 
+  ## Examples
+
+  ### Using the `:chunk` option
+
+    Using a `Kino.Frame`.
+
+      frame = Kino.Frame.new() |> Kino.render()
+
+      for word <- ["who", " *let*", " `the`", " **dogs**", " out"] do
+        text = Kino.Markdown.new(word, chunk: true)
+        Kino.Frame.append(frame, text)
+        Process.sleep(250)
+      end
+
+    Without using a `Kino.Frame`.
+
+      for word <- ["who", " *let*", " `the`", " **dogs**", " out"] do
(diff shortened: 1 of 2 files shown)


lib/kino/markdown.ex
```diff
@@ -51,6 +51,29 @@ defmodule Kino.Markdown do
       are merged into a single text. This is useful for streaming content.
       Defaults to `false`
 
+  ## Examples
+
+  ### Using the `:chunk` option
+
+    Using a `Kino.Frame`.
```
jonatanklosko, line 58:
```suggestion
  Using a `Kino.Frame`.
```

lib/kino/markdown.ex
```diff
@@ -51,6 +51,29 @@ defmodule Kino.Markdown do
...
+
+      frame = Kino.Frame.new() |> Kino.render()
+
+      for word <- ["who", " *let*", " `the`", " **dogs**", " out"] do
+        text = Kino.Markdown.new(word, chunk: true)
+        Kino.Frame.append(frame, text)
+        Process.sleep(250)
+      end
+
+    Without using a `Kino.Frame`.
```
jonatanklosko, line 68:
```suggestion
  Without using a `Kino.Frame`.
```

lib/kino/text.ex
```diff
@@ -37,6 +37,29 @@ defmodule Kino.Text do
       are merged into a single text. This is useful for streaming content.
       Defaults to `false`
 
+  ## Examples
+
+  ### Using the `:chunk` option
+
+    Using a `Kino.Frame`.
```
jonatanklosko, line 44:
```suggestion
  Using a `Kino.Frame`.
```

Issue #408, opened 2024-03-28
Custom Message in `Kino.Process.seq_trace`


mariari:
It would be nice to be able to supply a function to `Kino.Process.seq_trace`, that can help augment the naming process.

For the code I work on, I see a lot of `CAST: tuple`, making the use of the utility of `seq_trace` limited.

This can be solved in a few ways:

1. Letting users write custom pattern matches for the kinds of messages they experience (much like how [label_from_message](https://github.com/livebook-dev/kino/blob/v0.12.3/lib/kino/process.ex#L422) already matches certain patterns).
2. Letting users supply their own custom `label_from_message` function to completely override the process.

I might try my hand at `1.` and contribute upstream if the idea works out.

josevalim:
A custom label for messages function would be welcome indeed. Please do submit a PR :)

PR #411, opened 2024-03-28
Custom Label Logic

mariari:
This topic addresses #408

I added an extra argument to `seq_trace` and `render_seq_trace`, that accepts an extra argument to augment the labeling logic.

I have not tested the code yet, nor did I see any tests in the repo.

I hope it works.

I am now going to test this in a project I'm working on, and might add an extra commit that augments the example, showing off what the custom label can do.

I've also exposed `label_from_value` as I think it may add some value to people's custom's logic's

```elixir
  @spec foo(term()) :: label_response()
  defp foo(message) do
    case message do
      {:"$gen_call", _ref, {:special, _, msg}} -> "CALL: #{label_from_value(msg)}"
      {:"$gen_cast", _ref, {:special, _, msg}} -> "CAST: #{label_from_value(msg)}"
      _ -> :continue
    end
  end
```

is one such logic I used in mocking up what I wanted to do. I will get a better example as I make more examples.

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -486,7 +486,6 @@ defmodule Kino.Process do
   defp label_from_value(tuple) when is_tuple(tuple), do: "tuple"
   defp label_from_value(_), do: "term"
 
-  @spec label_from_options(keyword) :: (any() -> label_response)
   defp label_from_options(opts) do
     opts
     |> Keyword.get(:message_label, fn _message -> :continue end)
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -274,6 +274,33 @@ defmodule Kino.Process do
 
         Agent.stop(agent_pid)
       end)
+
+  Further if you are interested in custom labeling between messages
+  sent between processes, you can specify custom labels for the
+  messages you are interested in:
+
+      {:ok, agent_pid} = Agent.start_link(fn -> [] end)
+      Process.monitor(agent_pid)
+
+      Kino.Process.seq_trace(agent_pid, fn ->
+        1..2
+        |> Task.async_stream(
+          fn value ->
+            Agent.get(agent_pid, fn value -> value end)
+            100 * value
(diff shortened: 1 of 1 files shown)


mariari:
The code seems to work!

lib/kino/process.ex
```diff
@@ -224,6 +225,14 @@ defmodule Kino.Process do
   target argument can either be a single PID, a list of PIDs, or the atom `:all`
   depending on what messages you would like to retain in your trace.
 
+  ## Options
+
+    * `:label_function` - A function to help label message events. If
```
josevalim, line 230: Can we call it `:message_label`? We usually don't append the type to the option names.

I also don't think we should expose `label_from_value`. If you pass a custom function, you usually know better which values you are going to handle, no?

mariari: > Can we call it :message_label? We usually don't append the type to the option names.

Sure, let me change that now.

> I also don't think we should expose label_from_value. If you pass a custom function, you usually know better which values you are going to handle, no?

Most likely, I just wanted to preserve the initial message, as the code I'll be running it on just wraps the normal `GenServer.call` and `GenServer.cast` with some extra information to the given `GenServer`, meaning the default behavior is close to what I wanted but not quite.

mariari: Both complaints should be addressed now on the latest push

mariari:
I have also provided an example in the documentation now on how to run the new `message_label` option

lib/kino/process.ex
```diff
@@ -444,6 +486,12 @@ defmodule Kino.Process do
   defp label_from_value(tuple) when is_tuple(tuple), do: "tuple"
   defp label_from_value(_), do: "term"
 
+  @spec label_from_options(keyword) :: (any() -> label_response)
```
josevalim, line 489:
```suggestion
```

josevalim:

PR #412, opened 2024-04-02
Add function for getting client's user info

jonatanklosko:
See https://github.com/livebook-dev/livebook/pull/2540.

### DIFF
--- a/lib/kino/hub.ex
+++ b/lib/kino/hub.ex
@@ -31,4 +31,26 @@ defmodule Kino.Hub do
       {:error, _} -> %{type: :none}
     end
   end
+
+  @doc """
+  Returns user information for the given connected client id.
+
+  Note that this information is only available when the session uses
+  Livebook Teams hub, otherwise `:not_available` error is returned.
+
+  If there is no such connected client, `:not_found` error is returned.
+  """
+  @spec user_info(String.t()) :: {:ok, user_info()} | {:error, :not_found | :not_available}
+  def user_info(client_id) do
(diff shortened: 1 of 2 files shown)


lib/kino/hub.ex
```diff
@@ -31,4 +31,26 @@ defmodule Kino.Hub do
...
+  If there is no such connected client, `:not_found` error is returned.
+  """
+  @spec user_info(String.t()) :: {:ok, user_info()} | {:error, :not_found | :not_available}
+  def user_info(client_id) do
+    case Kino.Bridge.get_user_info(client_id) do
+      {:ok, user_info} ->
+        {:ok, user_info}
+
+      {:error, reason} when reason in [:not_found, :not_available] ->
+        {:error, reason}
```
josevalim, line 50: I am worried about this because, if we add new errors to Livebook in the future, old Kino versions will raise. Should we have {:error, reason} as a pass through and return something like `{:fail, reason}` when we actually want it to raise?

josevalim (approved):
One comment but your call :)

lib/kino/hub.ex
jonatanklosko: Yeah, I want to distinguish other request errors in a separate PR!

There is one place on the Livebook side where we return `{:error, :terminated}` and need to change to `{:request_error, :terminated}`, hence it's kinda breaking. It's fine though, because but it's not something we expected to happen and if it did, it wouldn't matter because of the consequences anyway :p

PR #413, opened 2024-04-03
Use more precise error tags when communicating with IO proxy

jonatanklosko:
🐈

### DIFF
--- a/lib/kino/bridge.ex
+++ b/lib/kino/bridge.ex
@@ -7,7 +7,7 @@ defmodule Kino.Bridge do
   # achieved via the group leader. For the implementation of
   # that group leader see Livebook.Evaluator.IOProxy
 
-  @type request_error :: :unsupported | :terminated
+  @type request_error :: {:request_error, :unsupported | :terminated}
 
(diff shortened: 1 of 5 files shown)


lib/kino/bridge.ex
```diff
@@ -257,10 +258,11 @@ defmodule Kino.Bridge do
 
     result =
       receive do
-        {:io_reply, ^ref, {:error, {:request, _}}} -> {:error, :unsupported}
-        {:io_reply, ^ref, {:error, :request}} -> {:error, :unsupported}
+        {:io_reply, ^ref, {:error, {:request, _}}} -> {:request_error, :unsupported}
+        {:io_reply, ^ref, {:error, :request}} -> {:request_error, :unsupported}
+        {:io_reply, ^ref, {:error, :terminated}} -> {:request_error, :terminated}
```
jonatanklosko, line 263: With this we don't actually need to change anything on Livebook side.

In retrospect, we should probably have a nested ok/error tuple to more clearly distinguish communication errors from expected errors, but it's fine.

PR #418, opened 2024-04-24
DataTable - export ordered data

cristineguadelupe:


jonatanklosko (approved):
This is a breaking change, but likely used only in `kino_explorer` and the next kino release is 0.13 anyway, so it's fine not to deprecate.

Issue #380, opened 2024-01-18
Feature request: Multi-select


Munksgaard:
The multi-select that is used in the data transformation smart cell when  "selecting" between outputs is very nice. Would it be possible to add something similar to `Kino.Control`?

josevalim:
Right now we are sticking with HTML supported elements, but this is totally something we can consider in the future. Thank you!

sergey-chechaev:
We really miss this feature. Maybe we'll reopen an issue

PR #426, opened 2024-05-11
Add Kino.Wizard POC

josevalim:


josevalim:
@hugobarauna check the docs for a proof of concept for you to try it out. I can flesh it out if we decide to include it in. The first argument to `init` has an `origin`, which can be used to identify the actual user.

Note: @jonatanklosko didn't like the function style and would prefer a handle_event style.

hugobarauna:
@josevalim, I think we can move forward with that. 👍

And if by _"jonatanklosko didn't like the function style and would prefer a handle_event style."_, it means that the API will be closer to a LiveView API; I agree with him.

While I was trying to use `Kino.Wizard`, I struggled a little with understanding how to use it. If it's closer to LiveView, it would be easier for me, as someone who has a basic understanding of how to write a LiveView.

sergey-chechaev:
Are there any plans for multi-select? Maybe there is a workaround, such as using a smart cell?

https://github.com/livebook-dev/kino/issues/380

josevalim:
Closing in favor of #489.

PR #431, opened 2024-05-23
Add `Kino.Proxy` module

aleDsz:


lib/kino/proxy.ex
```diff
@@ -0,0 +1,22 @@
...
+        |> Plug.Conn.send_resp(200, "used " <> conn.method <> " method")
+      end)
+  """
+
+  @doc """
+  Persists the function to be listened by the proxy handler.
+  """
+  @spec listen(atom(), module(), (Plug.Conn.t() -> Plug.Conn.t())) ::
+          DynamicSupervisor.on_start_child()
+  def listen(name \\ __MODULE__, mod \\ Livebook.Proxy.Handler, fun) when is_function(fun, 1) do
```
josevalim, line 19: I don't think we want those parameters.

We also most likely don't want to depend on Livebook.Proxy.Handler directly. Perhaps we add a function to Kino.Bridge called `fetch_proxy_handler_spec`, and we pass the `function` as argument?

jonatanklosko: The least coupling would be just registering the function using `Kino.Bridge` and be done, but given that we also want to use `Kino.start_child`, asking for the spec sounds sane!

lib/kino/proxy.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Kino.Proxy do
+  @moduledoc """
+  A kino for handling proxy requests from the host.
```
jonatanklosko, line 3: Let's not use the word kino, since it basically means an output :)

```suggestion
  Functionality for handling proxy requests forwarded from Livebook.
```

josevalim:
```suggestion
  Handle proxy requests from Livebook.
```

We will need to add extensive documentation on the URL format and include examples. :)

lib/kino/bridge.ex
```diff
@@ -250,6 +250,15 @@ defmodule Kino.Bridge do
     match?({:ok, _}, io_request(:livebook_get_evaluation_file))
   end
 
+  @doc """
+  Requests the child spec for proxy handler with the given function.
+  """
+  @spec get_proxy_handler_child_spec((Plug.Conn.t() -> Plug.Conn.t())) ::
+          Supervisor.child_spec() | {module(), term()} | module()
+  def get_proxy_handler_child_spec(fun) do
+    with {:ok, reply} <- io_request({:livebook_get_proxy_handler_child_spec, fun}), do: reply
```
josevalim, line 259: Doesn't this mean it can still return `{:error, ...}`? Should we handle it?

lib/kino/bridge.ex
```diff
@@ -250,6 +250,15 @@ defmodule Kino.Bridge do
     match?({:ok, _}, io_request(:livebook_get_evaluation_file))
   end
 
+  @doc """
+  Requests the child spec for proxy handler with the given function.
+  """
+  @spec get_proxy_handler_child_spec((Plug.Conn.t() -> Plug.Conn.t())) ::
+          Supervisor.child_spec() | {module(), term()} | module()
```
jonatanklosko, line 257: We are the ones controlling both sides, so we can use a more precise type, looking at the Livebook side that's `{module(), term()}` :D

lib/kino/bridge.ex
jonatanklosko, line 259:
```suggestion
    io_request({:livebook_get_proxy_handler_child_spec, fun})
```

and we can adjust the typespec to be `{:ok, {module(), term()}} | request_error()`. Then in `listen` we can match and raise an error accordingly, similarly to

https://github.com/livebook-dev/kino/blob/df73bed09b3363c40e35a47c8272c8d1e7642f2f/lib/kino/hub.ex#L33-L34

jonatanklosko (approved):

lib/kino/proxy.ex
```diff
@@ -0,0 +1,81 @@
...
+    * `:host`
+    * `:method`
+    * `:owner`
+    * `:port`
+    * `:remote_ip`
+    * `:query_string`
+    * `:path_info`
+    * `:scheme`
+    * `:script_name`
+    * `:req_headers`
```
jonatanklosko, line 33: These are all "request fields", so perhaps we should say that instead?

jonatanklosko:
I pushed some docs edits to avoid talking implementation details to the user. @josevalim feel free to finalize and merge :)

lib/kino/proxy.ex
```diff
@@ -0,0 +1,81 @@
+defmodule Kino.Proxy do
+  @moduledoc """
+  Functionality for handling proxy requests forwarded from Livebook.
+
+  Livebook proxies requests at the following paths:
+
+    * `/sessions/:id/proxy/*path` - a notebook sessions
```
jonatanklosko, line 7:
```suggestion
    * `/sessions/:id/proxy/*path` - a notebook session
```

lib/kino/proxy.ex
```diff
@@ -0,0 +1,81 @@
...
+    * `:script_name`
+    * `:req_headers`
+
+  > #### Plug {: .info}
+  >
+  > In order to use this feature, you need to add `:plug` as a dependency.
+
+  ## Examples
+
+  Using the proxy feature, you can use Livebook apps to build APIs.
```
jonatanklosko, line 41:
```suggestion
  Using the proxy feature, we can use Livebook apps to build APIs.
```

lib/kino/proxy.ex
josevalim: I don't think we should say which fields are set. For all intents and purpose, it should work as a regular webserver, and that's what a regular webserver would set.

lib/kino/proxy.ex
```diff
@@ -0,0 +1,81 @@
...
+    * `:host`
+    * `:method`
+    * `:owner`
+    * `:port`
+    * `:remote_ip`
+    * `:query_string`
+    * `:path_info`
+    * `:scheme`
+    * `:script_name`
+    * `:req_headers`
```
jonatanklosko, line 33:
```suggestion
```

lib/kino/proxy.ex
```diff
@@ -0,0 +1,81 @@
...
+  Functionality for handling proxy requests forwarded from Livebook.
+
+  Livebook proxies requests at the following paths:
+
+    * `/sessions/:id/proxy/*path` - a notebook session
+
+    * `/apps/:slug/:session_id/proxy/*path` - a specific app session
+
+    * `/apps/:slug/proxy/*path` - generic app path, only supported for
+      single-session apps
```
josevalim, line 12:
```suggestion
    * `/apps/:slug/proxy/*path` - generic app path, only supported for
      single-session apps. If the application has automatic shutdows
      enabled and it is not currently available, it will be automatically started
```

josevalim (approved):

Issue #430, opened 2024-05-23
Add ETS table rendering to Kino.Process.{sup_tree,app_tree}


akoutmos:
While working on the Elixir Patterns book I found it useful to include ETS tables in the `Kino.Process.sup_tree/1` rendering so readers could see the layout of process as well as any ETS tables associated with those processes. Something like so:

<img width="898" alt="image" src="https://github.com/livebook-dev/kino/assets/4753634/e802ebdc-18a6-4961-89bc-3bd1f94ff5a7">

If there is interest in adding this to Kino I can open up a PR with my work :).

akoutmos:
Thanks!

1. Correct. The solid line to the ETS table is the owning process and the dotted line is the heir process.
2. good idea....I didn't even think about that. Do you picture it using Mermaid subgraphs?

josevalim:
> Correct. The solid line to the ETS table is the owning process and the dotted line is the heir process.

I would invert the arrow then (perhaps annotate it with (heir) if possible.

> good idea....I didn't even think about that. Do you picture it using Mermaid subgraphs?

That would be the idea but I am not sure if it is better in practice.

akoutmos:
Sounds good. I'll put together a couple hard coded mocks and post them here with which route we want to go.

akoutmos:
Here are some hardcoded Mermaid diagrams for how we can render this:

Diagram without subgraph:
<img width="894" alt="image" src="https://github.com/livebook-dev/kino/assets/4753634/ec17e55c-54d5-4818-8339-a4a9d78daf04">

Diagram with subgraph:
<img width="896" alt="image" src="https://github.com/livebook-dev/kino/assets/4753634/8738b458-7d44-4ee4-aa95-0f429c24bc58">

We can also drop the arrows all together for ETS tables and label both owner and heir to be explicit
<img width="896" alt="image" src="https://github.com/livebook-dev/kino/assets/4753634/eedf2e0d-9be3-430d-9a46-58f04500593d">

Thoughts?

josevalim:
I think the original one is less cluttered. I would only add an arrow to heir, but pointing to the process instead. What are your thoughts? Any thoughts @hugobarauna?

akoutmos:
When the direction of the ETS owner edge changes, Mermaid shuffles the graph in a not so intuitive way:

<img width="889" alt="image" src="https://github.com/livebook-dev/kino/assets/4753634/50e74138-607d-4066-b636-f1330b5add9a">

[...]

PR #432, opened 2024-05-24
Render ets tables

akoutmos:
This PR addresses https://github.com/livebook-dev/kino/issues/430

By default, ETS table rendering is disabled and the rendering looks as it already does:
image

If you enable ETS table rendering you see the following:
image

At the moment this PR is lacking tests and could use a little refactoring, but I want to make sure the diagrams are right before writing the tests and performing last mile clean up.

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -727,9 +727,7 @@ defmodule Kino.Process do
   end
 
   defp graph_node(%{idx: idx, id: id, meta: %{protection: protection}, type: :ets}) do
-    """
-    #{idx}[("`#{module_or_atom_to_string(id)}\n**_#{protection}_**`")]:::ets
-    """
+    "#{idx}[(\"`#{module_or_atom_to_string(id)}\n**_#{protection}_**`\")]:::ets"
   end
 
   defp graph_node(%{idx: idx, pid: pid, type: type}) do
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -623,21 +623,39 @@ defmodule Kino.Process do
         end
       end)
 
-    ets_owner_map
-    |> Enum.reduce(supervision_tree_data, fn {ets_table_owner, table_info},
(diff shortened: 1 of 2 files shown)


lib/kino/process.ex
```diff
@@ -581,21 +586,98 @@ defmodule Kino.Process do
...
+    end
+  end
+
+  defp do_traverse_ets_tables(supervision_tree_data) do
+    active_ets_tables =
+      :ets.all()
+      |> Enum.map(fn table ->
+        table
+        |> :ets.info()
+        |> Map.new()
```
josevalim, line 620: It is most likely cheaper to lookup the information that you care about:

```suggestion
        %{owner: :ets.info(table, :owner), heir: :ets.info(table, :heir)}
```

You also need to consider the table can be deleted between `:ets.all` and getting its info. So you need to wrap it in a try catch:

```suggestion
      try do
        %{owner: :ets.info(table, :owner), heir: :ets.info(table, :heir)}
      rescue
        _ -> %{owner: nil, heir: nil}
      end
```

lib/kino/process.ex
```diff
@@ -581,21 +586,98 @@ defmodule Kino.Process do
...
+      end)
+
+    ets_heir_map =
+      active_ets_tables
+      |> Enum.reduce(%{}, fn table_info, acc ->
+        case Map.get(table_info, :heir) do
+          heir_pid when is_pid(heir_pid) -> Map.put(acc, heir_pid, table_info)
+          _ -> acc
+        end
+      end)
```
josevalim, line 639: You can fold the ets table lookup and both reduces into a single traversal:

    Enum.reduce(:ets.all, {%{}, %{}}, fn ... ->

You could also keep a single map if you prefer and keep the keys namespaced, such as `{:heir, pid}` and `{:owner, pid}`. It will be less data to pass around.

josevalim:

akoutmos:
Thanks for the review and the merge José!

PR #433, opened 2024-05-28
Add docs related to auth when using Kino.Proxy

hugobarauna:


lib/kino/proxy.ex
```diff
@@ -29,12 +29,10 @@ defmodule Kino.Proxy do
   Using the proxy feature, we can use Livebook apps to build APIs.
   For example, we could provide a data export endpoint:
 
-      data = <<...>>
-      token = "auth-token"
 
       Kino.Proxy.listen(fn
         %{path_info: ["export", "data"]} = conn ->
-          ["Bearer " <> ^token] = Plug.Conn.get_req_header(conn, "authorization")
+          data = "some data"
```
josevalim, line 35: In this case, using bearer tokens may be the best example, since tokens are most common for APIs (and Kino.Proxy would most likely be used for APIs?)

hugobarauna: Indeed.

I'll change from HTTP basic auth to HTTP Bearer.

hugobarauna: I thought about leaving this example as simple as possible, with no auth. And the auth would be below, inside the admonition block.

lib/kino/proxy.ex
```diff
@@ -46,8 +44,28 @@ defmodule Kino.Proxy do
...
+  > #### Authentication {: .warning}
+  >
+  > The paths exposed by `Kino.Proxy` don't use the authentication mechanisms
+  > defined in your Livebook instance.
+  >
+  > If you need to authenticate requests, you should
+  > implement your own authentication mechanism. Here's a simple example.
+  >
+  > ```elixir
+  > Kino.Proxy.listen(fn conn ->
```
hugobarauna, line 59: @josevalim what about this code example with a simple Bearer auth?

```elixir
Kino.Proxy.listen(fn conn ->
  api_token = "my-secret-api-token"

  with ["Bearer " <> client_token] <- Plug.Conn.get_req_header(conn, "authorization"),
       true <- api_token == client_token do
    Plug.Conn.send_resp(conn, 200, "hello")
  else
    _ ->
      conn
      |> Plug.Conn.put_resp_header("www-authenticate", "Bearer")
      |> Plug.Conn.send_resp(401, "Unauthorized")
  end
end)
```

hugobarauna (commented):
@josevalim maybe this version?

Although I'm not 100% satisfied with it. I accept suggestions.

lib/kino/proxy.ex
josevalim:
```
Kino.Proxy.listen(fn conn ->
  api_token = "my-secret-api-token"

  case Plug.Conn.get_req_header(conn, "authorization") do
    ["Bearer " <> ^api_token] ->
      Plug.Conn.send_resp(conn, 200, "hello")
    _ ->
      conn
      |> Plug.Conn.put_resp_header("www-authenticate", "Bearer")
      |> Plug.Conn.send_resp(401, "Unauthorized")
  end
end)

File: lib/kino/proxy.ex [hunk 3]
```diff
@@ -46,8 +44,28 @@ defmodule Kino.Proxy do
...
       end)
 
-  Once deployed as an app, the user would be able to export the data
+  Once deployed as an app, the API client would be able to export the data
   by sending a request to `/apps/:slug/proxy/export/data`.
+
+  > #### Authentication {: .warning}
+  >
+  > The paths exposed by `Kino.Proxy` don't use the authentication mechanisms
+  > defined in your Livebook instance.
```
jonatanklosko, line 53: @josevalim I started to wonder if we should put them in the `/public` namespace. The main downside is that it differs more from the base app/session path.

josevalim: We can always add more routes in the future. I think we have a good starting set. I can also see it being used for coordinating tasks inside their own infrastructure, but still not exposing it to the world.

lib/kino/proxy.ex
```diff
@@ -46,8 +44,30 @@ defmodule Kino.Proxy do
...
+  > implement your own authentication mechanism. Here's a simple example.
+  >
+  > ```elixir
+  > Kino.Proxy.listen(fn conn ->
+  >   api_token = "my-secret-api-token"
+  >
+  >   case Plug.Conn.get_req_header(conn, "authorization") do
+  >     ["Bearer " <> ^api_token] ->
+  >       Plug.Conn.send_resp(conn, 200, "hello")
+  >     _ ->
```
jonatanklosko, line 65:
```suggestion
  >       Plug.Conn.send_resp(conn, 200, "hello")
  >
  >     _ ->
```

hugobarauna:
@josevalim can I merge this?

josevalim:
@hugobarauna to merge this, you should either use Plug.BasicAuth, or keep the token aprpoach, but use `Plug.Crypto.secure_compare(user_token, expected_token)`., Then feel free to ship it!

hugobarauna:
@josevalim now it should be ready to merge. Please take a last look.

josevalim (approved):

Issue #440, opened 2024-06-10
Support custom data formatter for Kino.DataTable


kipcole9:
Currently `Kino.DataTable` formats data with either the `String.Chars` implementation or `inspect/2`.  However the default presentation isn't localizable or customisable - and since Livebooks can be targeted at a wider audience than just Elixir developers, an option to provide a formatter function would seem a good improvement to user experience.

The simplest option would be to add an optional formatter function to `Kino.DataTable.new/2`.  For example:
```elixir
Kino.DataTable.new(
  data,
  keys: [:registered_name, :initial_call, :reductions, :stack_size],
  formatter: &MyMod.formatter/2
)
```

The formatter function would receive the column value and the key so that formatting can be contextual. The default would be the current `value_to_string/1` function, slightly modified to accept (and ignore) the key.

If this idea is accepted I am very willing to create a PR. And of course if there is a better API thats good too.

josevalim:
:+1:. @cristineguadelupe @jonatanklosko, do you foresee any issues?

kipcole9:
I have made the necessary changes in [a fork of kino](https://github.com/kipcole9/kino) and implemented a livebook demonstrating an example using a formatter I built into a new release of [ex_cldr](https://hex.pm/packages/ex_cldr).  

[![Run in Livebook](https://livebook.dev/badge/v1/blue.svg)](https://livebook.dev/run?url=https%3A%2F%2Fraw.githubusercontent.com%2Felixir-cldr%2Fcldr%2Fv2.39.0%2Flivebook%2Fkino_data_table_formatting.livemd)

If you find the overall outcome acceptable I'll submit a PR. I'm unsure how to create tests for this change and I couldn't see any for the current implementation so I hope the example livebook is acceptable.

josevalim:
I dropped some nits, it looks great to me :)

josevalim:
Closing in favor of PR.

PR #441, opened 2024-06-11
Add a :formatter option to Kino.DataTable.new/2

kipcole9:
See the discusssion in #440.

Adds a `:formatter` option to `Kino.DataTable.new/2` which is a 2-arity function passed the `key` (column name) and cell value.

The special key `:__header__` is passed when formatting the column headings.

### DIFF
--- a/test/kino/data_table_test.exs
+++ b/test/kino/data_table_test.exs
@@ -277,12 +277,14 @@ defmodule Kino.DataTableTest do
 
   test "supports a formatter option" do
     entries = %{x: 1..3, y: [1.1, 1.2, 1.3]}
+
     formatter =
       fn
         :__header__, value -> {:ok, "h:#{value}"}
         :x, value when is_integer(value) -> {:ok, "x:#{value}"}
         _, _ -> :default
       end
+
     kino = Kino.DataTable.new(entries, keys: [:x, :y], formatter: formatter)
     data = connect(kino)
 


lib/kino/data_table.ex
```diff
@@ -83,7 +92,8 @@ defmodule Kino.DataTable do
   """
   def update(kino, tabular, opts \\ []) do
     {data_rows, data_columns, count, inspected} = prepare_data(tabular, opts)
-    Kino.Table.update(kino, {data_rows, data_columns, count, inspected})
+    formatter = Keyword.get(opts, :formatter, &value_to_string/2)
+    Kino.Table.update(kino, {data_rows, data_columns, count, inspected, formatter})
```
jonatanklosko, line 96: `update` is rather designed for updating data and I wouldn't expect formatter to change after the table is created. So let's support it only in `new` :)

lib/kino/data_table.ex
```diff
@@ -43,14 +43,23 @@ defmodule Kino.DataTable do
       data. Sorting requires traversal of the whole enumerable, so it
       may not be desirable for large lazy enumerables. Defaults to `true`
 
+   * `:formatter` - a 2-arity function that is used to format the data
+     in the table. The first parameter passed is the `key` (column name) and
+     the second is the value to be formatted. When formatting column headings
+     the key is the special value `:__header__`. Defaults to the builtin
+     formatter.
```
jonatanklosko, line 50: I would change the function to return `{:ok, string} | :default`. This way if someone wants to customize a specific value, like dates, they can do that without reimplementing the default logic for other values.

With that, the `:formatter` option wouldn't have a default value, if present we call the formatter first, otherwise we don't, in both cases we follow up with the default logic if applicable.

jonatanklosko:
Hey @kipcole9! I like the idea, just a small note on the API and we can ship it :D

jonatanklosko:
Ah and as for tests, please add one in `test/kino/data_table_test.exs`. For example here:

https://github.com/livebook-dev/kino/blob/93d042a7c72c77584004644391a52c7d2ef6f8ee/test/kino/data_table_test.exs#L249-L261

The `data: [...]` carries the already formatted values.

kipcole9:
@jonatanklosko, thanks for the review and great suggestions. I've made the requested changes, tests are passing. Now I just need to add a new test as you proposed.

kipcole9:
Test case added so I think this is now ready for the next review.

test/kino/data_table_test.exs
```diff
@@ -275,6 +275,21 @@ defmodule Kino.DataTableTest do
     })
   end
 
+  test "supports a formatter option" do
+    entries = %{x: 1..3, y: [1.1, 1.2, 1.3]}
+    formatter = &Kino.DataTable.Formatter.format/2
```
jonatanklosko, line 280: Let's keep it simple for the test and define a formatter inline:

```suggestion
    formatter =
      fn
        {:__header__, value} -> "h:#{value}"
        {:x, value} -> "x:#{value}"
        _ -> :default
      end
```

lib/kino/data_table.ex
```diff
@@ -279,17 +295,28 @@ defmodule Kino.DataTable do
...
+  end
+
+  defp value_to_string(key, value, formatter) do
+    case formatter.(key, value) do
+      {:ok, string} -> string
+      :default -> value_to_string(key, value)
+    end
+  end
+
+  defp value_to_string(_key, value) when is_atom(value), do: inspect(value)
```
jonatanklosko, line 309: We can remove the key now, since we never use it in the default logic :D

jonatanklosko:
@kipcole9 perfect, last two nitpicks!

lib/kino/data_table.ex
kipcole9: Done!

test/kino/data_table_test.exs
kipcole9: Done!

kipcole9:
Fixed as requested. Let me know if there's anything else?

test/kino/data_table_test.exs
```diff
@@ -275,6 +275,26 @@ defmodule Kino.DataTableTest do
...
 
+  test "supports a formatter option" do
+    entries = %{x: 1..3, y: [1.1, 1.2, 1.3]}
+    formatter =
+      fn
+        :__header__, value -> {:ok, "h:#{value}"}
+        :x, value when is_integer(value) -> {:ok, "x:#{value}"}
+        _, _ -> :default
+      end
+    kino = Kino.DataTable.new(entries, keys: [:x, :y], formatter: formatter)
```
jonatanklosko, line 286:
```suggestion
    entries = %{x: 1..3, y: [1.1, 1.2, 1.3]}

    formatter =
      fn
        :__header__, value -> {:ok, "h:#{value}"}
        :x, value when is_integer(value) -> {:ok, "x:#{value}"}
        _, _ -> :default
      end

    kino = Kino.DataTable.new(entries, keys: [:x, :y], formatter: formatter)
```

PR #442, opened 2024-06-11
Support for OTP 27 process labels

hugobarauna:
## Before example

CleanShot 2024-06-11 at 17 44 03

## After example

CleanShot 2024-06-11 at 17 44 34

URL with a notebook showing the feature being used in multiple contexts: https://gist.github.com/hugobarauna/6dd8b05b51d66bf313c3870c2bc428ed.

You can import that notebook, with Livebook running with Elixir 1.17 rc and see the diagrams with process labels.


lib/kino/process.ex
```diff
@@ -421,15 +430,37 @@ defmodule Kino.Process do
     {func_result, sequence_diagram}
   end
 
-  defp generate_participant_entry(pid, idx) do
+  defp get_label(pid) do
+    # :proc_lib.get_label/1 was added in OTP 27
```
josevalim, line 434:
```suggestion
  @compile {:no_warn_undefined, {:proc_lib, :get_label, 1}}

  defp get_label(pid) do
    # TODO: Remove :proc_lib.get_label/1 conditional from Erlang/OTP 27+
```

lib/kino/process.ex
```diff
@@ -315,6 +316,9 @@ defmodule Kino.Process do
 
   def seq_trace(trace_pids, trace_function, opts)
       when is_list(trace_pids) or trace_pids == :all do
+    # Set up the tracer responsible for storing calls to Process.set_label/1
+    {:ok, process_label_tracer_pid} = LabelTracer.start_link()
```
josevalim, line 320: Why do we need a process?

lib/kino/process.ex
```diff
@@ -740,13 +771,35 @@ defmodule Kino.Process do
...
+
+        {:registered_name, name} ->
+          module_or_atom_to_string(name)
       end
 
     "#{idx}(#{display}):::#{type}"
   end
 
+  defp format_for_mermaid_graph_node(pid, process_label) do
+    %{"pid" => pid_text} = Regex.named_captures(~r/#PID(?<pid>.*)/, inspect(pid))
```
josevalim, line 790: What if it is a named process? Or `self()`?

lib/kino/process.ex
```diff
@@ -740,13 +771,35 @@ defmodule Kino.Process do
 
     display =
       case process_info(pid, :registered_name) do
-        {:registered_name, []} -> inspect(pid)
-        {:registered_name, name} -> module_or_atom_to_string(name)
+        {:registered_name, []} ->
+          inspect(pid)
```
josevalim, line 775: ✂️

lib/kino/process.ex
```diff
@@ -421,15 +430,37 @@ defmodule Kino.Process do
...
+            "participant #{idx} AS #35;PID#{:erlang.pid_to_list(pid)};"
+
+          process_label ->
+            "participant #{idx} AS #{format_for_mermaid_participant_alias(pid, process_label)};"
+        end
     end
   end
 
+  defp format_for_mermaid_participant_alias(pid, process_label) do
+    %{"pid" => pid_text} = Regex.named_captures(~r/#PID(?<pid>.*)/, inspect(pid))
```
josevalim, line 458: Similar here, you can use :erlang.pid_to_list

test/kino/process_test.exs
```diff
@@ -114,13 +114,94 @@ defmodule Kino.ProcessTest do
...
+
+        {_func_result, diagram} = Kino.Process.seq_trace(trace_function)
+        diagram = mermaid(diagram)
+
+        assert diagram =~ ~r/participant 1 AS #{process_label}<br\/><0.\d+.0>;/
+      end
+    end
+  end
+
+  defmodule Ponger do
```
jonatanklosko, line 190: This module is not used :)

lib/kino/process.ex
jonatanklosko: We can use this:

```elixir
# TODO: remove once we require OTP 27
if Code.ensure_loaded?(:proc_lib) and function_exported?(:proc_lib, :get_label, 1) do
  defp get_label(pid), do: :proc_lib.get_label(pid)
else
  defp get_label(_pid), do: :undefined
end
```

lib/kino/process.ex
jonatanklosko: We can move the conditional to compilation time, no?

```elixir
if Code.ensure_loaded?(:proc_lib) and function_exported?(:proc_lib, :get_label, 1) do
  defp get_label(pid), do: :proc_lib.get_label(pid)
else
  defp get_label(_pid), do: :undefined
end
```

josevalim: Also works for me!

lib/kino/process.ex
hugobarauna: This was my second approach, the first didn't work.

In the first one, I was trying to get the process label from the process registry directly, using `:proc_lib.get_label/1`. But the problem was that by the time I was calling that function, the process I was trying to access the label from could have already died.

That happened in that simple example:

```elixir
parent = self()

Kino.Process.render_seq_trace(fn ->
  child =
    spawn(fn ->
      Process.set_label(:ponger)

      receive do
        :ping ->
          send(parent, :pong)
      end
    end)

  send(child, :ping)

  receive do
    :pong -> :ponged!
  end
end)
```

lib/kino/process.ex
hugobarauna: done in https://github.com/livebook-dev/kino/pull/442/commits/07f90b1ff2b13d1d1f65b488300ef525fb37da8c

lib/kino/process.ex
hugobarauna: Ahhh, forgot to delet the `inspect`, tks.

lib/kino/process.ex
hugobarauna: In the "ping <-> pong" example, it's not a race condition; it's deterministic

In this example, we won't be able to access the `:ponger` label using `:proc_lib.get_label/1` because we first run the tracer function (the anonymous function passed to `Kino.Process.render_seq_trace`), and only after that do we'd try to access the labels of the processes inside. By that time, the `:ponger` process is already dead (deterministically).

Since the "ping <-> pong" example is our (theoretically) "canonical example" for showing the sequence diagrams for rendering message passing, I'd prefer if our solution would work for showing process labels for that example.

What do you think?

Maybe I'm not seeing the downside of using the tracer here, since it's the first time I'm using it.

One downside I can see in the current implementation is that if the developer sets the process label using `:proc_lib.set_label/1` (from Erlang) instead of using the Elixir wrapper (`Process.set_label/1`), then the current code wouldn't get the process label. But we could easily address that by tracing that function call as well, if we want to."

jonatanklosko: Btw. we already have a tracker process, we could at least reuse it?

hugobarauna: > Btw. we already have a tracker process, we could at least reuse it?

That may be possible. I can investigate that path.

lib/kino/process.ex
wojtekmach: or let's use `inspect(pid)` and that's it?

Looking at the screenshot in PR description, before:

```
#PID<0.2234.0>
```

After:

```
:ponger
<0.2234.0>
```

I think doing this:

```
:ponger
#PID<0.2234.0>
```

doesn't hurt and it is super clear what the number is. Unless I'm missing something?

josevalim: Repeating PID every time is verbose IMO.

lib/kino/process.ex
```diff
@@ -421,15 +423,36 @@ defmodule Kino.Process do
     {func_result, sequence_diagram}
   end
 
-  defp generate_participant_entry(pid, idx) do
+  # :proc_lib.get_label/1 was added in OTP 27
```
jonatanklosko, line 426: Same in other places

```suggestion
  # TODO: use :proc_lib.get_label/1 once we require OTP 27
```

test/kino/process_test.exs
```diff
@@ -114,13 +114,94 @@ defmodule Kino.ProcessTest do
       assert content =~ "0(supervisor_parent):::root ---> 2(#{inspect(agent)}):::worker"
     end
 
+    # Process.set_label/1 was addeed in Elixir 1.17.0
```
jonatanklosko, line 117:
```suggestion
    # TODO: remove once we require Elixir v1.17.0
```

lib/kino/process.ex
```diff
@@ -421,15 +423,36 @@ defmodule Kino.Process do
...
+            "participant #{idx} AS #{format_for_mermaid_participant_alias(pid, process_label)};"
+        end
     end
   end
 
+  defp format_for_mermaid_participant_alias(pid, process_label) do
+    pid_text = :erlang.pid_to_list(pid) |> List.to_string()
+
+    participant_alias = "#{inspect(process_label)}<br/>#{pid_text}"
+    String.replace(participant_alias, "\"", "")
```
jonatanklosko, line 453: Should this also be `format_as_mermaid_unicode_text`? If so, let's move the replace into that function (my understanding is that we remove `"` because we want to enclose the text in `"`, so it's related).

lib/kino/process/tracer.ex
```diff
@@ -31,10 +31,30 @@ defmodule Kino.Process.Tracer do
       message: message
     }
 
-    {:noreply, [new_event | trace_events]}
+    trace_events = [new_event | trace_info.raw_trace_events]
+    process_labels = update_process_labels(trace_info.process_labels, from, to)
```
jonatanklosko, line 35:
```suggestion

    process_labels = 
      trace_info.process_labels
      |> put_new_label(from)
      |> put_new_label(to)
```

lib/kino/process/tracer.ex
```diff
@@ -31,10 +31,30 @@ defmodule Kino.Process.Tracer do
...
 
   def handle_info(_ignored_event, trace_events) do
     {:noreply, trace_events}
   end
+
+  defp update_process_labels(process_labels, pid1, pid2) do
+    process_labels
+    |> put_new_label(pid1)
+    |> put_new_label(pid2)
+  end
```
jonatanklosko, line 48:
```suggestion
```

lib/kino/process.ex
hugobarauna: That `String.replace` is to remove the quotes that comes when we `inspect(...)` a string.

Otherwise, if the process label has a string set as label, the label in the diagram would include the quotes, like that:

CleanShot 2024-06-13 at 10 45 07

The other `format_as_mermaid_unicode_text` is there because if one uses unicode chars inside the process label, mermaid needs to enclose the text with `"` to be able to display unicode text: https://mermaid.js.org/syntax/flowchart.html#unicode-text

lib/kino/process.ex
hugobarauna: @josevalim ok, it should be fixed now. 

Still there's a race condition, but I think it's ok.

For example, with the new solution, I was able to reproduce the race condition with that example:

```elixir
parent = self()

child =
  spawn(fn ->
    Process.set_label(:ponger)

    receive do
      :ping ->
       # theres's a race condition here, without that sleep
       # when the tracer tries to get the process label
       # the ponger process is already dead
       Process.sleep(1) 
       send(parent, :pong)
    end
  end)

Kino.Process.render_seq_trace(fn ->
  send(child, :ping)

  receive do
    :pong -> :ponged!
  end
end)
```

Also, the old solution had a bug, as you mentioned, it could not get process labels set outside the traced function. Now that's working: 

CleanShot 2024-06-13 at 10 53 02

hugobarauna:
@josevalim @jonatanklosko, I think this should be good to go. Can you please take a last look?

lib/kino/process.ex
```diff
@@ -421,15 +423,36 @@ defmodule Kino.Process do
...
```
jonatanklosko, line 453: I see, in that case I would do the replace on inspect specifically for clarity:

```suggestion
    label = process_label |> inspect() |> String.replace(~s{"}, "")
    "#{label}<br/>#{pid_text}"
```

lib/kino/process/tracer.ex
```diff
@@ -31,10 +31,28 @@ defmodule Kino.Process.Tracer do
...
+
+    {:noreply, %{trace_info | raw_trace_events: trace_events, process_labels: process_labels}}
   end
 
   def handle_info(_ignored_event, trace_events) do
     {:noreply, trace_events}
   end
+
+  defp put_new_label(process_labels, pid) do
+    Map.put_new(process_labels, pid, get_label(pid))
```
josevalim, line 49:
```suggestion
    Map.put_new_lazy(process_labels, pid, fn -> get_label(pid) end)
```
This way we only compute the label if necessary.

josevalim (approved):
One comment and we can ship it!!!!

lib/kino/process.ex
```diff
@@ -421,15 +423,36 @@ defmodule Kino.Process do
...
+            "participant #{idx} AS #{format_for_mermaid_participant_alias(pid, process_label)};"
+        end
     end
   end
 
+  defp format_for_mermaid_participant_alias(pid, process_label) do
+    pid_text = :erlang.pid_to_list(pid) |> List.to_string()
+
+    label = process_label |> inspect() |> String.replace(~s{"}, "")
+    "#{label}<br/>#{pid_text}"
```
wojtekmach, line 453: nitpick, I'd change to:

```elixir
    label = process_label |> inspect() |> String.replace(~s{"}, "")
    "#{label}<br/>#{:erlang.pid_to_list(pid)}"
```

I tried pushing to main but I don't have access. :D No big deal anyway of course!

PR #446, opened 2024-06-20
Raise if Plug is missing

josevalim:
@jonatanklosko I wonder how hard it would be for us to add a special exception that automatically injects a button that allows us to click it and the dependency is automatically added. :)

### DIFF
--- a/lib/kino/proxy.ex
+++ b/lib/kino/proxy.ex
@@ -80,6 +80,16 @@ defmodule Kino.Proxy do
   """
   @spec listen((Plug.Conn.t() -> Plug.Conn.t())) :: DynamicSupervisor.on_start_child()
   def listen(fun) when is_function(fun, 1) do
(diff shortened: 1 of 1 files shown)


wojtekmach:
Like [`Plug.Exception.actions/1`](https://hexdocs.pm/plug/Plug.Exception.html#actions/1)?

jonatanklosko (approved):

jonatanklosko:
FTR we discussed we could add `Kino.MissingPackagesError` (similar to a few other exceptions we have), but it would be nice to have something more generic, because this kind of exception is more common to be raised by packages with optional dependencies (also, having a dedicated exception just for this one case is a bit of an overkill). We could have a heuristic matching on `RuntimeError` message, but that's not reliable. Finally, we could have a more specific error in Elixir, but it's weird that the package requirement is really about Mix. For now we are sticking to the regular exception.

PR #448, opened 2024-06-20
Accept module plugs in Kino.Proxy.listen/1

wojtekmach:
An example would be:

```elixir
defmodule AppRouter do
  use Plug.Router

  plug :match
  plug :dispatch

  get "/hello" do
    send_resp(conn, 200, "world")
  end

  match _ do
    send_resp(conn, 404, "oops")
  end
end

Kino.Proxy.listen(AppRouter)
```

### DIFF
--- a/lib/kino/bridge.ex
+++ b/lib/kino/bridge.ex
@@ -259,12 +259,12 @@ defmodule Kino.Bridge do
   end
 
   @doc """
-  Requests the child spec for proxy handler with the given plug.
+  Requests the child spec for proxy handler with the given function.
   """
-  @spec get_proxy_handler_child_spec(Kino.Proxy.plug()) ::
+  @spec get_proxy_handler_child_spec((Plug.Conn.t() -> Plug.Conn.t())) ::
           {:ok, {module(), term()}} | request_error()
-  def get_proxy_handler_child_spec(plug) do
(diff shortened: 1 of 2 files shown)


lib/kino/bridge.ex
```diff
@@ -259,12 +259,12 @@ defmodule Kino.Bridge do
...
   @doc """
-  Requests the child spec for proxy handler with the given function.
+  Requests the child spec for proxy handler with the given plug.
   """
-  @spec get_proxy_handler_child_spec((Plug.Conn.t() -> Plug.Conn.t())) ::
+  @spec get_proxy_handler_child_spec(Kino.Proxy.plug()) ::
           {:ok, {module(), term()}} | request_error()
-  def get_proxy_handler_child_spec(fun) do
-    io_request({:livebook_get_proxy_handler_child_spec, fun})
+  def get_proxy_handler_child_spec(plug) do
```
josevalim, line 266: I don't think we need to change here. We can convert the plugs into functions in Kino.Proxy and keep this contract small.

wojtekmach: Good call, I’ll update soon.

josevalim (approved):

jonatanklosko (approved):

PR #449, opened 2024-06-22
Show worker/supervisor ids in app supervision graph

DiaanEngelbrecht:
Hi there!

TLDR; I added labels to the process nodes to make it more clear what the processes represent.

I was using `Kino.Process` in Livebook to try and figure out how some elements of the Membrane RTC Engine work. I won't go into the details, but suffice to say this is a foreign code base to me, and it wasn't clear from the get go:

* What the supervision tree looks like.
* How it the internal genservers relate to each other.
* How that ties into my phoenix app.

So I was poking around blindly until I got my RTC Engine's process's graphed with Kino. This is what it looked like for reference:
without_initial_call

Cool! So now I get to see the process layout, but I'm still blind as to what all this actually is. I still don't know what Membrane is doing. To get a bit more insight, changed the label for the PID node to also print out the node's id and got this:

with_id

This helped me relate my WebRTC endpoints with pid's and I also really like the insight it gives when you plot a Phoenix app's supervision tree.

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -751,7 +751,7 @@ defmodule Kino.Process do
     "#{idx}[(\"`#{module_or_atom_to_string(id)}\n**_#{protection}_**`\")]:::ets"
   end
 
-  defp graph_node(%{idx: idx, pid: pid, type: type}) do
+  defp graph_node(%{idx: idx, id: id, pid: pid, type: type}) do
     type =
       if idx == 0 do
         :root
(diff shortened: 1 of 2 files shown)


josevalim (approved):

jonatanklosko (approved):

lib/kino/process.ex
```diff
@@ -763,7 +763,7 @@ defmodule Kino.Process do
       case process_info(pid, :registered_name) do
         {:registered_name, []} ->
           case get_label(pid) do
-            :undefined -> inspect(pid)
+            :undefined -> format_for_mermaid_graph_node(pid, id)
             process_label -> format_for_mermaid_graph_node(pid, process_label)
```
jonatanklosko, line 767: For the origin process we currently show `:root`, which can be confusing, so I would skip it

```suggestion
            :undefined ->
              if idx == 0 do
                inspect(pid)
              else
                # Use worker/supervisor id as label
                format_for_mermaid_graph_node(pid, id)
              end

            process_label ->
              format_for_mermaid_graph_node(pid, process_label)
```

josevalim:

PR #452, opened 2024-07-03
Update docs on nested Kino.start_child/1

jonatanklosko:
🐈

### DIFF
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -428,7 +428,7 @@ defmodule Kino do
   > or `c:Kino.JS.Live.init/2`. If you do that, starting the process
   > will block forever.
   >
-  > Creating many kinos uses `start_child/1` underneath, which means
+  > On creation, many kinos use `start_child/1` underneath, which means
   > that you cannot use functions such as `Kino.DataTable.new/1` in
   > `c:GenServer.init/1`. If you need to do that, you must either
   > create the kinos beforehand and pass in the `GenServer` argument,
--- a/lib/kino.ex
+++ b/lib/kino.ex
@@ -428,9 +428,11 @@ defmodule Kino do
   > or `c:Kino.JS.Live.init/2`. If you do that, starting the process
   > will block forever.
   >
-  > Note that creating many kinos uses `start_child/1` underneath,
-  > hence the same restriction applies to starting those. See
-  > `c:Kino.JS.Live.init/2` for more details.
+  > Creating many kinos uses `start_child/1` underneath, which means
+  > that you cannot use functions such as `Kino.DataTable.new/1` in
+  > `c:GenServer.init/1`. If you need to do that, you must either
+  > create the kinos beforehand and pass in the `GenServer` argument,
+  > or create them in `c:GenServer.handle_continue/2`.
   """
   @spec start_child(
           Supervisor.child_spec()
--- a/lib/kino/js/live.ex
+++ b/lib/kino/js/live.ex
@@ -201,6 +201,9 @@ defmodule Kino.JS.Live do
   Invoked when the server is started.
 
   See `c:GenServer.init/1` for more details.
+
+  If you want to create other kinos on initialization, see the
+  limitations described in `Kino.start_child/1`.
   """
   @callback init(arg :: term(), ctx :: Context.t()) ::
               {:ok, ctx :: Context.t()} | {:ok, ctx :: Context.t(), opts :: keyword()}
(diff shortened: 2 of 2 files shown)


lib/kino/js/live.ex
```diff
@@ -202,44 +202,8 @@ defmodule Kino.JS.Live do
...
-  >
-  >     defmodule KinoDocs.Custom do
-  >       def new() do
-  >         Kino.JS.Live.new(__MODULE__, {})
-  >       end
-  >
-  >       @impl true
-  >       def init({}, ctx) do
-  >         frame = Kino.Frame.new()
-  >         {:ok, assign(ctx, frame: frame)}
```
jonatanklosko, line 220: This example is confusing, we create a frame in `Kino.JS.Live`, but there is no obvious way how we would use that frame (since `Kino.JS.Live` itself is already about managing JS rendering). If we want a frame to render updates into, as we do in the dbg pipeline, it is actually more natural to create the frame upfront and pass as an argument.

I think it's more about creating kinos like `Kino.DataTable.new()` and rendering into a provided frame, and it's also more about `GenServer`, because it's a rare case for `Kino.JS.Live`.

lib/kino.ex
```diff
@@ -428,9 +428,11 @@ defmodule Kino do
   > or `c:Kino.JS.Live.init/2`. If you do that, starting the process
   > will block forever.
   >
-  > Note that creating many kinos uses `start_child/1` underneath,
-  > hence the same restriction applies to starting those. See
-  > `c:Kino.JS.Live.init/2` for more details.
+  > Creating many kinos uses `start_child/1` underneath, which means
```
josevalim, line 431:
```suggestion
  > On creation, many kinos uses `start_child/1` underneath, which means
```

josevalim (approved):

lib/kino.ex
```diff
@@ -428,9 +428,11 @@ defmodule Kino do
...
   > will block forever.
   >
-  > Note that creating many kinos uses `start_child/1` underneath,
-  > hence the same restriction applies to starting those. See
-  > `c:Kino.JS.Live.init/2` for more details.
+  > On creation, many kinos use `start_child/1` underneath, which means
+  > that you cannot use functions such as `Kino.DataTable.new/1` in
+  > `c:GenServer.init/1`. If you need to do that, you must either
+  > create the kinos beforehand and pass in the `GenServer` argument,
+  > or create them in `c:GenServer.handle_continue/2`.
```
hugobarauna, line 435: Ohhh, didn't know it was possible to do that inside `handle_continue` ! 💡

timgremore: @hugobarauna I've been trying to find an example of this. Do you happen to know of an example smartcell that uses `handle_continue/2` to start a child kino?

The issue I'm running into is with [Kino..SmartCell.Server.validate_init_opts!/2](https://github.com/livebook-dev/kino/blob/1f63f9075c93922e12ce33079144fa5c987d2a51/lib/kino/smart_cell/server.ex#L118) requiring init options.

jonatanklosko: @timgremore I don't think `Kino.start_child` is relevant for smart cell implementation. The idea is that you use `Kino.start_child` to start processes in *cell code* that gets evaluated, so that Livebook knows that it should kill the process before the cell is evaluated again. On the other hand, a smart cell starts and runs in the background, regardless of what gets evaluated (it's just a server process for driving the UI and code generation). If you need the smart cell to start a background process, you can just use `start_link`.

> Do you happen to know of an example smartcell that uses handle_continue/2 to start a child kino?

By "child kino" do you mean like `Kino.DataTable.new(...)` or similar? Those kinos are generally supposed to be used for cell outputs only. What is your use case? :)

timgremore: @jonatanklosko I think I follow the distinction you're explaining. My application involves a `:gen_statem` implementation that I'm representing in UI. I attempted to start my state machine with something like (this the child kino I referenced above):

```elixir
Kino.start_child!(%{
  start: {MyStateServer, :start_link, []},
  id: MyStateServer
})
```

I had thought `handle_continue/2` would be an appropriate spot for that call. But I understand your point about that being more suitable for cell code.

Seems `start_link` is more fitting for what I'm trying to achieve but, if I understand correctly, I'll need something else to start `:gen_statem`. Is that correct?

jonatanklosko: If I understand correctly that you are defining a smart cell and starting the `:gen_statem` server from the smart cell, you can just do `MyStateServer.start_link([])` in the smart cell init and it should be all good.

timgremore: That's it! Thank you @jonatanklosko for helping me sort that out

PR #456, opened 2024-07-11
Don't show :undefined id when rendering supervision trees

hugobarauna:
## Before

image

## After

image

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -764,7 +764,7 @@ defmodule Kino.Process do
         {:registered_name, []} ->
           case get_label(pid) do
             :undefined ->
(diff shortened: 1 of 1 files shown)


lib/kino/process.ex
```diff
@@ -764,7 +764,7 @@ defmodule Kino.Process do
         {:registered_name, []} ->
           case get_label(pid) do
             :undefined ->
-              if idx == 0 do
+              if idx == 0 || id == :undefined do
```
jonatanklosko, line 767:
```suggestion
              if idx == 0 or id == :undefined do
```

jonatanklosko (approved):

PR #461, opened 2024-07-26
Add kino_orb for huge Network Graphs

marcnnn:
Hey

First Thanks for Kino and Livebook it's amazing!

I was looking into bigger Graphs and how to work with them in Livebook and found https://github.com/memgraph/orb.

So I just tried it in with Kino.JS.

Do you think it would fit into the Kino?
or:
Should I create a separate Repo for that?
Would you like to create a Repo under the livebook-dev Org?

Thanks
Marc


josevalim:
A separate `kino_orb` package is the way to go, that's also how we handle things like `VegaLite`, and we want to welcome community packages, so under your own name or org is best. Thank you for your contribution!

PR #463, opened 2024-08-03
Fix various typos and improve language

preciz:
Thank you for reviewing.

### DIFF
--- a/test/kino/video_test.exs
+++ b/test/kino/video_test.exs
@@ -4,7 +4,7 @@ defmodule Kino.VideoTest do
   describe "new/2" do
     test "raises an error for a non-image MIME type" do
       assert_raise ArgumentError,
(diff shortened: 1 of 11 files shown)


test/kino/video_test.exs
```diff
@@ -4,15 +4,15 @@ defmodule Kino.VideoTest do
   describe "new/2" do
     test "raises an error for a non-image MIME type" do
       assert_raise ArgumentError,
-                   ~s{expected video type to be either :mp4, :ogg, :avi, :wmv, :mov, or an video MIME type string, got: "application/json"},
+                   ~s{expected video type to be either :mp4, :ogg, :avi, :wmv, :mov, or a video MIME type string, got: "application/json"},
```
jonatanklosko, line 7: We need to adjust the raising code also :D

jonatanklosko (approved):
Thanks, just one detail and we can merge :)

test/kino/video_test.exs
preciz: True!

PR #470, opened 2024-09-19
Call off start_child when the group leader terminates

jonatanklosko:
Example

```elixir
defmodule Slow do
  use GenServer

  def start_link(time) do
    GenServer.start_link(__MODULE__, time, name: __MODULE__)
  end

  def init(time) do
    Process.sleep(time)
    {:ok, time}
  end
end

Kino.start_child({Slow, 10_000})
```

When we "Evaluate" and "Stop" (which terminates the evaluator), the child still ends up starting. Consequently, if we try to evaluate again (which starts a new evaluator), we get `:already_started`, because the process is still under the kino supervision tree.

### DIFF
--- a/lib/kino/terminator.ex
+++ b/lib/kino/terminator.ex
@@ -32,14 +32,13 @@ defmodule Kino.Terminator do
       if pid do
         terminator = cross_node_name()
 
-        with :ok <- Kino.Bridge.reference_object(pid, parent),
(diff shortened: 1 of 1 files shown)


josevalim (approved):

hugobarauna:
> we get :already_started

I got this yesterday while starting a FLAME Pool. Felt weird, but it worked at the end.

Maybe it was related to this bug. If yes, hopefully fixed now. 😊

PR #473, opened 2024-10-03
Proposal: Add shortcut to nothing in Kino.Shorts

christhekeele:
As I make more sophisticated Kino UIs:

- I find myself using `Kino.Shorts` more to streamline large complicated blocks of Elixir Kino UI code
- I find myself invariably terminating those blocks with `Kino.nothing()` since I've taken control of the rendering process

In the spirit of `Kino.Shorts` reducing the times you have to type `Kino`, I thought a shortcut function might make sense. For a use-case extracted from a pattern I've used several times now, see the second example in the function docs.

The name I've chosen here (`output_nothing`) is debatable:

- On one hand it is two characters longer than the function it is aliasing
- On the other hand it reads more clearly and declaratively than scattering kind of context-less `Kino.nothing()`s at the end of code blocks

### DIFF
--- a/lib/kino/shorts.ex
+++ b/lib/kino/shorts.ex
@@ -9,6 +9,44 @@ defmodule Kino.Shorts do
 
   """
 
+  @doc """
+  Returns a special value that results in no visible output.
+
+  It is a wrapper around `Kino.nothing/0`.
+
+  ## Examples
+
+  Supress an otherwise verbose cell's output:
+
+      import Kino.Shorts
+
+      resp = Req.get!("https://example.org")
+      output_nothing()
+
+  For convenience, you can pipe anything into `output_nothing()`
+  to terminate a pipeline of Kinos that control their own rendering:
+
+      import Kino.Shorts
+
+      frame = frame() |> Kino.render
+      frame |> Kino.Frame.render(markdown("> ***Sumbit an event***"))
+
+      Kino.Control.form(
+        [name: Kino.Input.text("Event Name", default: "default")],
+        submit: "Render Event"
+      )
+      |> Kino.render
+      |> Kino.listen(fn
+        event ->
+          Kino.Frame.clear(frame)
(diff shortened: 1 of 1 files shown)


josevalim (commented):
I would go with `nothing()`. I am not sure I like nothing/1 because it gives the impression something is done with the argument. A `nothing()` alone in the last line is clearer IMO.

christhekeele:
I do feel like including a verb in the function name improves the readability (here and in most function names). `Kino.nothing()` is a noun because it is returning the special `t:Kino.nothing` atom, but once aliased just calling `nothing()` tells me nothing about what the function call accomplishes.

>  I am not sure I like nothing/1 because it gives the impression something is done with the argument.

Perhaps a different verb would make you feel more comfortable with it, say `ignore_output`?

josevalim:
The issue is that it is inconsistent with everything else. `audio`, `mermaid`, etc are all outputs, we don't label any of them.

> Perhaps a different verb would make you feel more comfortable with it, say ignore_output/1?

Not really, it has no relationship with the input in any case. :)

christhekeele:
> The issue is that it is inconsistent with everything else. audio, mermaid, etc are all outputs, we don't label any of them.

I see what you mean, they are all nouns too. Despite that, I still like the readable verb form and `/1` arity, since the fact that the function does not return something kino-renderable makes it kind of a non-entity in Kino terms. While `Kino.render(nothing())` reads nicely, providing the verb context, just `nothing()` does not to me.

If you're not sold on verb-name or arity-1, though, I think it's fine to close this PR, as the exact same thing as your proposed alterations can be accomplished with more clarity via `import Kino, only: [nothing: 0]`

josevalim:
I am ok with adding `nothing/0` as a shortcut, as the functions here are only meant to be shortcut, not new functionality. But if you don't see a value in that, we can close this.

PR #476, opened 2024-10-08
DataTable - support setting number of rows

### LINKED ISSUE #475: [Feature Request] Allow passing number of rows to show in Kino.DataTable
Hi,

Thanks for the project. It is mind-blowingly fantastic! :)

I have a little nit about DataTable. I am using it for data set that is exactly 14 rows long.
When DataTable is first rendered, the default option is to show first 10 rows + pagination.
So, after each reevaluation, I need to click "show 20" to see everything.
(note that show: :all also does not solve the issue, because it adds scroll instead of pagination).
I would like to be able to pass initial values to `show` and `page` when initializing `DataTable`, e.g.

```
Kino.DataTable.new(data, keys: [:first_row, :second_row], show: 20, page: 1)
```

Alternatively, it could be nested:

```
Kino.DataTable.new(data, keys: [:first_row, :second_row], display_options: [show: 20, page: 1])
```

WDYT?

cristineguadelupe:
Closes #475.


jonatanklosko (approved):

lib/kino/data_table.ex
```diff
@@ -172,9 +175,9 @@ defmodule Kino.DataTable do
   end
 
   @impl true
-  def init({data_rows, data_columns, count, name, sorting_enabled, inspected, formatter}) do
+  def init({data_rows, data_columns, count, name, sorting_enabled, inspected, formatter, num_rows}) do
```
jonatanklosko, line 178:
```suggestion
  def init(
        {data_rows, data_columns, count, name, sorting_enabled, inspected, formatter, num_rows}
      ) do
```

lib/kino/data_table.ex
```diff
@@ -172,9 +175,11 @@ defmodule Kino.DataTable do
   end
 
   @impl true
-  def init({data_rows, data_columns, count, name, sorting_enabled, inspected, formatter}) do
+  def init(
+        {data_rows, data_columns, count, name, sorting_enabled, inspected, formatter, num_rows}
+      ) do
     features = Kino.Utils.truthy_keys(pagination: true, sorting: sorting_enabled)
-    info = %{name: name, features: features}
+    info = %{name: name, features: features, num_rows: num_rows}
```
jonatanklosko, line 182: Just so we are consistent with the typespec and never pass `nil` :)

```suggestion
    info = %{name: name, features: features}
    info = if(num_rows, do: Map.put(info, :num_rows, num_rows), else: info)
```

jonatanklosko:
Thanks! 🐈

PR #490, opened 2025-02-11
Add style to Kino.Text

josevalim:


lib/kino/text.ex
```diff
@@ -37,6 +40,11 @@ defmodule Kino.Text do
       are merged into a single text. This is useful for streaming content.
       Defaults to `false`
 
+    * `:style` - a keyword list or map of pairs HTML styles, such as
```
jonatanklosko, line 43: The typespec above only supports keywordlist, so perhaps let's do only that? Also HTML style is an implementation detail, so probably makes more sense to mention CSS instead.

```suggestion
    * `:style` - a keyword list of CSS attributes, such as
```

jonatanklosko (approved):
Just one comment and ship it!

PR #491, opened 2025-02-12
Use Kino.Render.Any.to_livebook/1

josevalim:
Otherwise folks may copy and paste the contents
of this file, accidentally using Kino.Output's private
API. Relying on Kino.Render.Any is ok because that's
a public fallback.

Other options to consider: use Kino.Text |> Kino.Render.to_livebook,
although that's more verbose.


jonatanklosko (approved):

jonatanklosko:
Good call :)

Issue #494, opened 2025-02-26
Removed the deprecated `:export_info_string` option from `Kino.JS.new/3` causes error


deadego:
https://github.com/livebook-dev/kino_vega_lite/blob/main/lib/kino/vega_lite.ex#L47-L53 is still calling with the recently removed options  calling the error below. 

```
[error] ** (ArgumentError) unknown keys [:export_key, :export_info_string] in [export: #Function<3.65502277/1 in Kino.VegaLite.static/1>, export_info_string: "vega-lite", export_key: :spec], the allowed keys are: [:export]
    (elixir 1.18.2) lib/keyword.ex:362: Keyword.validate!/2
    (kino 0.15.0) lib/kino/js.ex:504: Kino.JS.new/3
    (kino_vega_lite 0.1.11) lib/kino_vega_lite.ex:3: Kino.Render.VegaLite.to_livebook/1
    lib/livebook/runtime/evaluator/formatter.ex:62: Livebook.Runtime.Evaluator.Formatter.to_output/1
    lib/livebook/runtime/evaluator.ex:494: Livebook.Runtime.Evaluator.continue_do_evaluate_code/6
    lib/livebook/runtime/evaluator.ex:335: Livebook.Runtime.Evaluator.loop/1
    (stdlib 6.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

https://github.com/livebook-dev/kino/commit/cf2cdd6f511adaa10308ec9dba193b8ed128c008#diff-f36022180d3057a0a81deca673c7e7a11d0d07d0b4242e376d08614c239282ddR504-R505

hugobarauna:
I was able to replicate the bug using that notebook source:

````livemd
# Untitled notebook

```elixir
Mix.install([
  {:vega_lite, "~> 0.1.11"},
  {:kino_vega_lite, "~> 0.1.13"},
  {:kino, "~> 0.15.0"}
])
```

## Section

```elixir
alias VegaLite, as: Vl

data = [
  %{"a" => "A", "b" => 28},
  %{"a" => "B", "b" => 55}
]

Vl.new(width: 400, height: 300)
|> Vl.data_from_values(data)
|> Vl.mark(:bar)
|> Vl.encode_field(:x, "a", type: :nominal, axis: [label_angle: 0])
|> Vl.encode_field(:y, "b", type: :quantitative)
```

````

[...]

PR #495, opened 2025-02-27
Ignore :export_key passed to Kino.JS.new/3 instead of raising

deadego:
add back and simply ignore :export_key for now. #494


jonatanklosko (approved):
I was silly to miss this, thank you!

PR #497, opened 2025-03-04
Improves UI issues when rendering a Kino.DataTable

hugobarauna:
## Context
When rendering a `Kino.Datable`, I encountered two issues:

### 1. Table header does not fit within the width

CleanShot 2025-03-04 at 15 05 01

### 2. Fonts are applied only after hovering

https://github.com/user-attachments/assets/c9cc0328-556a-49ec-a973-e1f3cfd5aeda

This PR resolves problem 1 for Chrome and Firefox.

## Before

https://github.com/user-attachments/assets/0771b0c6-a902-4894-b1f6-1b7b641adc0a

## After

https://github.com/user-attachments/assets/3b9a70c4-5d9b-498b-9056-6ad0fb3c3e62


assets/packs/data_table/main.js
```diff
@@ -4,19 +4,66 @@ import "@glideapps/glide-data-grid/dist/index.css";
...
+  fontPreloader.setAttribute("aria-hidden", "true");
+  fontPreloader.style.cssText = `
+    position: absolute;
+    visibility: hidden;
+    pointer-events: none;
+    left: -9999px;
+  `;
+  fontPreloader.innerHTML =
+    "<span style=\"font-family:'JetBrains Mono'\">Font preload</span>" +
+    "<span style=\"font-family:'Inter'\">Font preload</span>";
```
jonatanklosko, line 36: We need to have a sample text for every font width, otherwise only the width 400 is loaded for each.

assets/packs/data_table/main.js
```diff
@@ -4,19 +4,66 @@ import "@glideapps/glide-data-grid/dist/index.css";
...
+
+  try {
+    if (document.fonts && document.fonts.ready) {
+      await Promise.race([
+        document.fonts.ready,
+        new Promise((resolve) => setTimeout(resolve, FONT_LOAD_TIMEOUT)),
+      ]);
+    } else {
+      await new Promise((resolve) => setTimeout(resolve, FONT_LOAD_TIMEOUT));
+    }
```
jonatanklosko, line 51: We shouldn't do the timeout.

When `document.fonts.ready` is available, we should just wait however long it takes, it may very well take a second, and since this is only the first render, it should be fine. Otherwise, we would wait 0.5s and then run into the same issues this solves.

When `document.fonts.ready` is not available, we would *always* wait the timeout, even if the fonts are actually cached, so for most renders it would slow things down unnecessarily. Looks like it's well supported anyway.

assets/packs/data_table/Skeleton.js
```diff
@@ -0,0 +1,64 @@
...
+    <div class="w-full">
+      <table class="w-full border-separate border-spacing-0 mt-2 animate-pulse">
+        <thead>
+          <tr class="bg-slate-50 h-[1.5rem]">
+          ${createHeaderCell("w-12")}
+          ${createHeaderCell("w-32")}
+          ${createHeaderCell("w-48")}
+          ${createHeaderCell("w-20")}
+          </tr>
+        </thead>
```
jonatanklosko, line 15: I think it's fine to skip the header, the difference from rows is very subtle, and since it's a placeholder it's barely noticeable anyway.

jonatanklosko (commented):
Looks good to me, I have a few comments, but I experimented with it and applied the comments already. Let me know if it works as expected for you :)

jonatanklosko:
Thanks!

hugobarauna:
@jonatanklosko and whenever we can release a new kino, so that users could benefit from that improvement, it would be great.

Also, now that this works for Chrome and Firefox, do you have ideas on how to make it work for Safari?

I tried a few different things but could not make it work properly.

PR #502, opened 2025-05-03
Display image urls as images in table cells

etareduction:
URL's that point to images are now detected by MIME type using HEAD request, and then displayed in the table cell.

image


jonatanklosko:
Hey!

Unfortunately, I don't think we should be doing a synchronous request before rendering the table, and I would prefer to avoid extra dependencies unless strictly necessary. Also, the user may still prefer to render those as actual URLs.

If we are to support it, I think it should be explicit and opt-in. It can be an option like this:

```elixir
Kino.DataTable.new(..., types: %{key1: "image"})
```

And type must match one of the [supported ones](https://github.com/livebook-dev/kino/blob/dbe4d8a0845941c000407761e4572842495a33d6/lib/kino/table.ex#L33-L38).

etareduction:
@jonatanklosko

Hello. I thought about this problem for a while, and got some ideas like what you're saying but didn't want to change the API in any significant way like adding new configuration options without consulting with someone. 

What you've shown seems alright to me, i can try re-implementing it this way. 

---

> And type must match one of the [supported ones](https://github.com/livebook-dev/kino/blob/dbe4d8a0845941c000407761e4572842495a33d6/lib/kino/table.ex#L33-L38).

Also i have updated the types list doc already in this PR.

jonatanklosko:
Sounds good!

> Also i have updated the types list doc already in this PR.

Oh yes, what I meant is that we would need to check the `:types` option to make sure it valid types are provided.

etareduction:
@jonatanklosko
Sorry for the delay, was busy at my day job last week. Does this last commit look alright to you?

assets/packs/data_table/App.js
```diff
@@ -275,6 +277,14 @@ export function App({ ctx, data }) {
     [content],
   );
 
+  const mapCellData = (cellKind, cellData) => {
+    if (cellKind === GridCellKind.Image && !Array.isArray(cellData)) {
```
jonatanklosko, line 281: The data we pass is always a string, so I think we don't need to check for the array case.

```suggestion
    if (cellKind === GridCellKind.Image) {
```

assets/packs/data_table/App.js
```diff
@@ -283,10 +293,12 @@ export function App({ ctx, data }) {
         ? content.data[col][row]
         : content.data[row][col];
 
+      const mappedCellData = mapCellData(kind, cellData)
```
jonatanklosko, line 296: From the naming it's not clear why we map anything. For clarity, I would change the name above `cellData -> formattedValue` and here `mapCellData(kind, cellData) -> getCellData(kind, formattedValue)`.

lib/kino/table.ex
```diff
@@ -92,6 +94,8 @@ defmodule Kino.Table do
       This works the same as `Kino.JS.new/3`, except the function
       receives the state as an argument
 
+    * `:types` - a map of type overrides for the columns.
+      The keys are the column names and the values are the types to be used in place of inferred ones.
```
jonatanklosko, line 98:
```suggestion
    * `:types` - a map of display type overrides for the columns. The keys
      are the column names and each value must be one of `t:type/0`. By
      default the types are inferred from the data values
```

The user needs to know what types are available and I would link to `t:type/0` to keep it in a single place.

lib/kino/table.ex
```diff
@@ -115,9 +119,14 @@ defmodule Kino.Table do
...
 
   @impl true
-  def init({module, init_arg}, ctx) do
+  def init({module, init_arg, types}, ctx) do
     {:ok, info, state} = module.init(init_arg)
 
+    if types != nil and not Enum.all?(types, fn {_, value} -> value in @types end) do
+      raise ArgumentError,
+            "Invalid column types were provided: #{inspect(types)}. Valid types are: #{inspect(@types)}"
+    end
```
jonatanklosko, line 128: Let's move this check to `new`, so it is validated before the process starts and should have a clearer stacktrace.

assets/packs/data_table/App.js
etareduction: When writing that i thought that maybe we could use the fact that our frontend data grid library supports displaying multiple images in a cell, and provide an API for this later on in the elixir side. But yeah that would require re-thinking that system where currently we only pass strings to the frontend.

assets/packs/data_table/App.js
etareduction: Fixed

lib/kino/table.ex
etareduction: Good catch! Fixed

assets/packs/data_table/App.js
etareduction: On second thought, yeah, it might be unnecessary for now.

lib/kino/table.ex
```diff
@@ -100,7 +105,14 @@ defmodule Kino.Table do
         fn ctx -> export.(ctx.assigns.state) end
       end
 
-    Kino.JS.Live.new(__MODULE__, {module, init_arg}, export: export)
+    types = opts[:types]
+
+    if types != nil and not Enum.all?(types, fn {_, value} -> value in @types end) do
+      raise ArgumentError,
+            "Invalid column types were provided: #{inspect(types)}. Valid types are: #{inspect(@types)}"
+    end
```
jonatanklosko, line 113:
```suggestion
    types = opts[:types] || %{}

    for {_key, value} <- types, value not in @types do
      raise ArgumentError,
            "got invalid column type: #{inspect(types)}, expected one of: #{inspect(@types)}"
    end
```

jonatanklosko:
@etareduction looks good to me! It would be good to add a test for this as well. It can be similar to this:

https://github.com/livebook-dev/kino/blob/8d23de16072c68c4eac5ce6934e6f003fdad821d/test/kino/data_table_test.exs#L155-L167

We can pass the `:types` option and assert that the `columns: [...]` sent to the client have the expected type field.

PR #510, opened 2026-03-13
Do not override dbg_callback implicitly

jonatanklosko:
Currently we override `:dbg_callback` compile env at runtime, which causes issues such as https://github.com/elixir-lang/elixir/issues/15178. This PR removes it in favour of https://github.com/livebook-dev/livebook/pull/3152.

### DIFF
--- a/lib/kino/application.ex
+++ b/lib/kino/application.ex
@@ -4,9 +4,6 @@ defmodule Kino.Application do
   use Application
 
   def start(_type, _args) do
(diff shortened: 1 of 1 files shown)


jonatanklosko:
@josevalim I am thinking we wait for Elixir v1.20, then ship new Livebook with that version in both Desktop and Docker, and then also ship Kino. wdyt?

josevalim (approved):

josevalim:
Sounds good!

PR #511, opened 2026-04-17
Update docs accordingly with Livebook changes

aleDsz:


lib/kino/workspace.ex
```diff
@@ -22,6 +26,9 @@ defmodule Kino.Workspace do
   Note that `:started_by` information is only available for multi-session
   apps when the app uses a Livebook Teams workspace.
 
+  If there is set of parameters used to start the session, `:session_params` map is returned.
+  This is only available for multi-session apps.
```
jonatanklosko:
```suggestion
  Multi-session apps can be started with `lb_*` query parameters. Those parameters
  are available in `:session_params` map.
```

jonatanklosko (approved):

PR #512, opened 2026-05-24
Handle genserver replies in Kino.Process.seq_trace

laszlokorte:
Currently `Kino.Process.seq_trace` does not show GenServer reply messages.
This is caused by the `to` field of a reply being a `Reference` instead of a `pid`.

This causes the replies to be discarded by `normalize_message` during mermaid generation.


josevalim:

PR #513, opened 2026-06-02
Implements row action for data table

aleDsz:
### Demo
https://github.com/user-attachments/assets/e86fe724-4c54-4696-95d2-93f91b6caf0e


lib/kino/data_table.ex
```diff
@@ -43,27 +43,31 @@ defmodule Kino.DataTable do
...
+    * `:formatter` - a 2-arity function that is used to format the data
+      in the table. The first parameter passed is the `key` (column name) and
+      the second is the value to be formatted. When formatting column headings
+      the key is the special value `:__header__`. The formatter function must
+      return either `{:ok, string}` or `:default`. When the return value is
+      `:default` the default data table formatting is applied.
 
     * `:num_rows` - the number of rows to show in the table. Defaults to `10`.
 
+    * `:actions` - the list of actions per row. Defaults to empty list.
```
josevalim: We need a bit more docs here. Perhaps a separate section? We need to document what are the list elements and how do you process them when an action is clicked.

josevalim (approved):

lib/assets/data_table/build/chunk-SOT7EBMF.js.map
jonatanklosko, line 1: @aleDsz when running `npm run dev`, the assets are built with sourcemaps. Before committing/merging we need `npm run build`, so that we only store the assets minified and without sourcemaps. I've just rebuilt in 1cb2c8c :)

test/kino/data_table_test.exs
```diff
@@ -344,4 +344,25 @@ defmodule Kino.DataTableTest do
...
+
+  test "supports actions column" do
+    entry = Enum.at(@people_entries, 0)
+    action = %{tag: :my_clicked_action, label: "Click me!"}
+
+    kino = Kino.DataTable.new(@people_entries, actions: [action])
+    myself = self()
+
+    :erlang.trace(Process.whereis(Kino.SubscriptionManager), true, [:receive, tracer: self()])
+    :erlang.trace_pattern(:receive, [], [])
```
jonatanklosko, line 356: Is there a reason we need this? The implementation details should not matter, as long as we receive the event correctly (which the last assertion tests).

lib/assets/data_table/build/chunk-SOT7EBMF.js.map
aleDsz: Oh, I forgot to run this command, sorry.

test/kino/data_table_test.exs
aleDsz: To ensure we can receive the message using `Kino.listen/2`

jonatanklosko: The last assertion:

```elixir
assert_receive {tag, ^entry} when tag == action.tag
```

already covers that, no? If `Kino.listen` doesn't work, we don't receive that message.

aleDsz: Unfortunately, no. If you remove the `:erlang.trace/3` it doesn't emit the event to `Kino.listen/2`, which doesn't allow us to run the last assertion

jonatanklosko: Ah I see, we need to ensure listen subscribes before pushing the event, we do the same elsewhere already, got it :)

PR #515, opened 2026-08-31
Add `Kino.Process.flush/0` and `render_flush/0`

LostKobrakai:
This is a simple counterpart to `IEx.Helpers.flush` to show the render out the messages of the current mailbox.

### DIFF
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -865,7 +865,7 @@ defmodule Kino.Process do
     Kino.Layout.tabs(do_flush(1, []))
   end
 
-  def do_flush(index, acc) do
+  defp do_flush(index, acc) do
     receive do
       msg -> do_flush(index + 1, [{"Message ##{index}", msg} | acc])
     after
--- a/lib/kino/process.ex
+++ b/lib/kino/process.ex
@@ -860,14 +860,32 @@ defmodule Kino.Process do
 
   @doc """
   Generates a tab interface of flushed messages.
+
+  ## Options
+
+    * `:layout` - Selects how to render the flushed messages. The
+      value can either `:list` or `:tabs`. Defaults to `:list`.
+
   """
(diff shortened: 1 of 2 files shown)


jonatanklosko (approved):
Looks good! The only thing I wonder is whether we should use `grid` instead of `tabs`, so all messages are visible without clicking around? @josevalim preference?

LostKobrakai:
I did consider grid, but for my usecase tab is actually better. I guess we could make it configurable – or just return a list and leave it to users, but that would make `render_flush` more tricky.

For configurable I'd be with you that grid might be the better default.

jonatanklosko:
Yeah, let's add `layout: :list | :tabs` option to both functions with `:list` as default, and then we can ship it!

lib/kino/process.ex
```diff
@@ -857,4 +857,50 @@ defmodule Kino.Process do
...
+      msg -> do_flush([msg | acc])
+    after
+      0 -> :lists.reverse(acc)
+    end
+  end
+
+  @doc """
+  Renders a tab interface of flushed messages.
+
+  This function renders tab interface of flushed messages like `flush/0` with the
```
jonatanklosko:
```suggestion
  Clears out process messages and renders all of them.

  This function renders flushed messages like `flush/0` with the
```

lib/kino/process.ex
```diff
@@ -857,4 +857,50 @@ defmodule Kino.Process do
   defp process_info(pid, spec) do
     :erpc.call(node(pid), Process, :info, [pid, spec])
   end
+
+  @doc """
+  Generates a tab interface of flushed messages.
```
jonatanklosko:
```suggestion
  Clears out process messages and arranges them into a list or tabs layout.
```

Clears out process messages and renders all of them.

PR #517, opened 2026-09-02
Non started processes do not have their ids quoted when needed

LostKobrakai:
This caused syntax errors for `sup_tree` for me.


josevalim:

